PR #11, opened 2013-01-23
Button action event added

apsavin:
What about unified action event for buttons?

### DIFF
--- a/desktop.blocks/button/button.js
+++ b/desktop.blocks/button/button.js
@@ -80,8 +80,13 @@ BEM.DOM.decl('button', /** @lends Button.prototype */ {
 
         'pressed' : function(modName, modVal) {
 
-            this.isDisabled() || this.trigger(modVal == 'yes' ? 'press' : 'release');
-
+            if (modVal == 'yes') {
+		this.trigger('press');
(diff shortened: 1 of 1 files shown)


desktop.blocks/button/button.js
```diff
@@ -80,8 +80,13 @@ BEM.DOM.decl('button', /** @lends Button.prototype */ {
 
         'pressed' : function(modName, modVal) {
 
-            this.isDisabled() || this.trigger(modVal == 'yes' ? 'press' : 'release');
```
tadatuta, line 83: we should check if the button isn't disabled

desktop.blocks/button/button.js
```diff
@@ -80,8 +80,13 @@ BEM.DOM.decl('button', /** @lends Button.prototype */ {
...
-
+            if (modVal == 'yes') {
+		this.trigger('press');
```
tadatuta, line 84: indentation

dfilatov:
Can you provide a real case where unified actions is useful?

apsavin:
Well, I think, we can just use 'release' button event.

PR #149, opened 2013-08-12
Possibility to specify complex data for input's value (v2 backport)

### LINKED ISSUE #148: input: Could not specify localized input as value from BEMJSON
```
{ block: 'input', mods: { size: 's' }, name: 'my-name', value: {
    block: 'i-bem', elem: 'i18n',
    keyset: 'my-keyset', key: 'my-key'
}, content: { elem: 'input' } }
```

narqo:
\cc @veged 

Could you review BEMHTML template, 'cause I still got few experience with new JS-syntax?

Fix #148 (v2)

### DIFF
--- a/common.blocks/input/input.bemhtml
+++ b/common.blocks/input/input.bemhtml
@@ -1,11 +1,27 @@
 block('input')(
     tag()('span'),
+
     js()(true),
+
     def()(function() {
         var ctx = this.ctx;
         ctx.id = ctx.id || this.generateId();
+        ctx.val = apply('value', { ctx: ctx.val });
         applyNext({ _input: ctx })
     }),
+
+    mode('value')(
+        match()(function() { return '' }),
+        match(this.ctx)(function() {
+            var value = [];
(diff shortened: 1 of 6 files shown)


common.blocks/input/input.bemhtml
```diff
@@ -1,11 +1,27 @@
 block('input')(
     tag()('span'),
+
     js()(true),
+
     def()(function() {
         var ctx = this.ctx;
         ctx.id = ctx.id || this.generateId();
+        ctx.val = apply('value', { ctx: ctx.val });
```
veged: I suggest simplify separate mode and applying process to:

``` js
var val = ctx.val;
ctx.val = this._.isSimple(val) ? val : BEMHTML.apply(val);
```

if we really need to extend something with calculating value, we can introduce separate templates mode later

narqo: «Ох тыж ёжик...» o_O``

veged: ?

narqo:
See #195

PR #233, opened 2013-11-15
Update index page

### LINKED ISSUE #232: Update index page

mishanga:
Close #232


mishanga:
@narqo take a look plz

desktop.pages/.bem/level.js
```diff
@@ -10,7 +10,10 @@ exports.getConfig = function() {
         bundleBuildLevels: [
```
narqo: This could (should?) be moved to `make.js`: see `BundleNode#getLevels()`

desktop.pages/index/index.bemjson.js
```diff
@@ -11,138 +10,132 @@
...
-            },
-            { tag: 'br' },
-            {
-                block: 'attach',
-                mods: { size: 's', disabled: 'yes' },
-                content: [
+            [
+                'Link',
+                {
+                    block: 'link',
```
narqo: Code-style is not consistent. A half of a code is written with "bem-core"-style (e.g. `key : val`), but some parts written like this `key: val`,

desktop.pages/.bem/level.js
mishanga: @narqo can you please help me with this? It's too hard for me :(

PR #243, opened 2013-11-23
Event 'release' added for block 'button'

apsavin:


apsavin:
There is no need for this event, we have 'click' instead.

Issue #255, opened 2013-12-18
button: Single-line comments are removed from the generated CSS


gela-d:
Single-line comments are removed from the generated CSS

PR #256, opened 2013-12-18
button: Single-line comments are removed from the generated CSS

gela-d:
Issue - https://github.com/bem/bem-components/issues/255


gela-d:
@mishanga

mishanga:
@gela-d so what? :) What's the problem?

PR #283, opened 2014-01-31
Add scaffolding for specs for popup

narqo:
- add bacis specs infrastructure for popup
- fix `csscomb` lint
- fix `deps`

\cc @dfilatov @veged


veged:
here is just "one-and-half" tests ;-) let's add some more

dfilatov:
@narqo, could you rebase your pull request?

narqo:
`nom run unit` in still failing with `touch-*` but it's ok ;)

PR #301, opened 2014-02-19
Specify proper dimentions for html and body tags in popup's tests

### LINKED ISSUE #300: popup: Fix tests for touch levels
`popup_autoclosable` tests doesn't work, because `body` has no height dimensions and one could not make proper "outside" click to close popup.

narqo:
fix #300


narqo:
Tests are failing because of bem/bem-core#396

PR #326, opened 2014-03-18
Move common parts of _focused/_disabled behavior to "control" block

### LINKED ISSUE #308: Think about mixin for focusable blocks
For example in `input`, `button` and `menu`.

narqo:
**WIP! DO NOT MERGE**

\cc @dfilatov 

The idea is to have a base block `base-control`, which implements common "focusable + disablable", well,.. interface ;)

The `base-control` itself never should be instantiate. It should only be used as a `baseBlock`. 

It provides reactions for `_focused`, `_disabled` modifiers, some protected methods like `_onFocus`, `_onBlur`, which are used as events handlers in _control's inheritors_. Also there are some common API methods: 
- `getName()` to get control's "name" (_returns an empty string if control doesn't have one_)
- `getVal()` to get its "value" (_returns an empty string if control doesn't have one_)

As a side notes: _I'am still not sure about the solution. It doesn't look "natural", but a least it allows to move a lot of copy and paste parts from other components, like `button`, `link`, etc. @dfilatov I need your opinion_ ;)

closes #308

### DIFF
--- a/common.blocks/base-control/base-control.js
+++ b/common.blocks/base-control/base-control.js
@@ -1,0 +1,108 @@
+/** @module base-control */
+
+modules.define(
+    'base-control',
+    ['i-bem__dom', 'dom', 'next-tick'],
+    function(provide, BEMDOM, dom, nextTick) {
+
+/**
+ * @exports
+ * @class base-control
+ * @abstract
+ * @bem
+ */
+provide(BEMDOM.decl(this.name, /** @lends base-control.prototype */{
+    beforeSetMod : {
+        'focused' : {
+            'true' : function() {
+                return !this.hasMod('disabled');
+            }
+        }
+    },
+
+    onSetMod : {
+        'js' : {
+            'inited' : function() {
+                this._focused = dom.containsFocus(this.elem('control'));
+                this._focused?
+                    // if control is already in focus, we need to set focused mod
+                    this.setMod('focused') :
+                    // if block already has focused mod, we need to focus control
+                    this.hasMod('focused') && this._focus();
+            }
+        },
+
+        'focused' : {
+            'true' : function() {
+                this._focused || this._focus();
+            },
+
+            '' : function() {
+                this._focused && this._blur();
+            }
+        },
+
+        'disabled' : {
+            '*' : function(modName, modVal) {
+                this.elem('control').prop(modName, !!modVal);
+            },
+
+            'true' : function() {
+                this.delMod('focused');
+            }
+        }
+    },
+
+    /**
+     * Returns name of control
+     * @returns {String}
+     */
+    getName : function() {
+        return this.elem('control').attr('name') || '';
+    },
+
+    /**
+     * Returns control value
+     * @returns {String}
+     */
+    getVal : function() {
+        return this.elem('control').val();
+    },
+
+    _onFocus : function() {
+        this._focused = true;
+        this.setMod('focused');
+    },
+
+    _onBlur : function() {
+        this._focused = false;
+        this.delMod('focused');
+    },
+
+    _focus : function() {
+        dom.isFocusable(this.elem('control')) && this.elem('control').focus();
+    },
+
+    _blur : function() {
+        this.elem('control').blur();
+    }
+}, /** @lends base-control */{
+    live : function() {
+        this
+            .liveBindTo('control', 'focusin', this.prototype._onFocus)
+            .liveBindTo('control', 'focusout', this.prototype._onBlur);
+
+        var focused = dom.getFocused();
+        if(focused.hasClass(this.buildClass('control'))) {
+            var _this = this; // TODO: https://github.com/bem/bem-core/issues/425
+            nextTick(function() {
+                if(focused[0] === dom.getFocused()[0]) {
+                    var block = focused.closest(_this.buildSelector());
+                    block && block.bem(_this.getName());
+                }
+            });
+        }
+    }
+}));
+
+});
--- a/common.blocks/button/button.js
+++ b/common.blocks/button/button.js
@@ -1,19 +1,23 @@
+/**
+ * @module button
+ */
+
 modules.define(
     'button',
-    ['i-bem__dom', 'jquery', 'dom', 'functions'],
-    function(provide, BEMDOM, $, dom, functions) {
+    ['i-bem__dom', 'base-control', 'jquery', 'dom', 'functions'],
+    function(provide, BEMDOM, BaseControl, $, dom, functions) {
 
 var KEY_CODE_SPACE = 32,
     KEY_CODE_ENTER = 13;
 
-provide(BEMDOM.decl(this.name, {
+/**
+ * @exports
+ * @class button
+ * @augments base-control
+ * @bem
+ */
+provide(BEMDOM.decl({ block : this.name, baseBlock : BaseControl }, /** @lends button.prototype */{
     beforeSetMod : {
-        'focused' : {
-            'true' : function() {
-                return !this.hasMod('disabled');
-            }
-        },
-
         'pressed' : {
             'true' : function() {
                 return !this.hasMod('disabled') || this.hasMod('togglable');
@@ -24,7 +28,7 @@ provide(BEMDOM.decl(this.name, {
     onSetMod : {
         'js' : {
             'inited' : function() {
-                this._focused = false;
+                this.__base.apply(this, arguments);
                 this._refocusOnBlur = false;
             }
         },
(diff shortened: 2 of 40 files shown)


common.blocks/control/control.js
```diff
@@ -0,0 +1,64 @@
...
+        },
+
+        'disabled' : {
+            'true' : function() {
+                this.delMod('focused');
+            }
+        }
+    },
+
+    _getFocusable : function() {
```
narqo: We need this because some controls, like `input`, use there elements as `focusable`. So they should override this method.

narqo:
I purposely haven't touched original block's specs about there `focused/disabled` behavior.

narqo:
- [x] Rename to `base-control`
- [x] Add `__control` which will catch focus (instead of `_getFocusable()`)
- [x] Add `__control` to blocks which doesn't have it
- [x] Add `live`-section, like this https://github.com/bem/bem-components/blob/issues/307%40v2/common.blocks/menu/menu.js#L144

narqo:
Think about port `disable`-state to `base-control`

narqo:
@dfilatov @veged please, remind me to discuss this on our next meeting :))

common.blocks/base-control/base-control.js
```diff
@@ -0,0 +1,87 @@
...
+        var controlPrototype = this.prototype;
+        this
+            .liveBindTo('control', 'focusin', controlPrototype._onFocus)
+            .liveBindTo('control', 'focusout', controlPrototype._onBlur);
+
+        var focused = dom.getFocused();
+        if(focused.hasClass(this.buildClass('control'))) {
+            var _this = this; // TODO: https://github.com/bem/bem-core/issues/425
+            nextTick(function() {
+                focused[0] === dom.getFocused()[0] && focused.bem(_this.getName());
```
narqo: add test about destruction in same tick with initialization

common.blocks/base-control/base-control.js
```diff
@@ -0,0 +1,87 @@
...
+            },
+
+            '' : function() {
+                this._focused && this._blur();
+            }
+        },
+
+        'disabled' : {
+            'true' : function() {
+                this.delMod('focused');
```
narqo, line 51: add `dom.isDisablable` (bem/bem-core#488) and do `if(dom.isDisablable(this.elem('control'))) this.elem('control').prop('disabled', true)`

common.blocks/button/button.js
```diff
@@ -1,16 +1,23 @@
...
 modules.define(
     'button',
-    ['i-bem__dom', 'jquery', 'dom'],
-    function(provide, BEMDOM, $, dom) {
-
-provide(BEMDOM.decl(this.name, {
+    ['i-bem__dom', 'base-control', 'jquery', 'dom', 'functions'],
+    function(provide, BEMDOM, BaseControl, $, dom, functions) {
+
+var KEY_CODE_SPACE = 32,
```
narqo, line 10: moving of `desktop.blocks/button/button.js` to `common.blocks/button/button.js` should be done as separate issue (#359)

common.blocks/button/button.js
```diff
@@ -97,28 +119,15 @@ provide(BEMDOM.decl(this.name, {
...
-        this.domElem.focus();
-    },
-
-    _blur : function() {
-        this.domElem.blur();
-    }
+    _doAction : functions.noop
 }, {
     live : function() {
+        var buttonPrototype = this.prototype;
```
narqo: `ptp`

common.blocks/checkbox/checkbox.js
```diff
@@ -52,39 +29,13 @@ provide(BEMDOM.decl(this.name, {
...
-            .liveBindTo('focusin', function() {
-                this._onFocus();
-            })
-            .liveBindTo('focusout', function() {
-                this._onBlur();
-            })
-            .liveBindTo('change', function() {
-                this._onChange();
-            });
+        this.liveBindTo('change', this.prototype._onChange);
```
narqo: `this.liveBindTo('control', 'change'...)`

common.blocks/input/input.js
```diff
@@ -72,35 +57,10 @@ provide(BEMDOM.decl(this.name, {
      */
     getName : function() {
```
narqo: move to `base-control`

common.blocks/menu/menu.js
```diff
@@ -29,29 +22,25 @@ provide(BEMDOM.decl(this.name, /** @lends menu.prototype */{
...
                 this.unbindFromDoc('keydown', this._onKeyDown);
-                this._focused && this._blur();
                 this._hoveredItem && this._hoveredItem.delMod('hovered');
+                this.__base.apply(this, arguments);
             }
         },
 
         'disabled' : function(modName, modVal) {
+            this.__base.apply(this, arguments);
             this.domElem.attr('tabindex', modVal? -1 : 0);
```
narqo: move to `base-control`
use `this.elem('control')` instead of `this.domElem`

common.blocks/button/button.js
```diff
@@ -97,28 +119,15 @@ provide(BEMDOM.decl(this.name, {
...
-            .liveBindTo('pointerpress', function() {
-                this._onPointerPress();
-            })
-            .liveBindTo('pointerclick', function(e) {
-                this._onPointerClick(e);
-            });
+            .liveBindTo('control', 'pointerpress', buttonPrototype._onPointerPress)
+            .liveBindTo('control', 'pointerclick', buttonPrototype._onPointerClick);
+
+        return this.__base.apply(this, arguments);
```
narqo, line 137: can use chain for `.__base` (everywhere)

common.blocks/radio/radio.js
```diff
@@ -1,19 +1,21 @@
+/**
+ * @module radio
+ */
+
 modules.define(
     'radio',
-    ['i-bem__dom', 'jquery', 'dom', 'radio-option'],
-    function(provide, BEMDOM, $, dom) {
+    ['i-bem__dom', 'base-control', 'jquery', 'dom', 'radio-option'],
+    function(provide, BEMDOM, BaseControl, $, dom) {
```
narqo: should not be augmented from `base-control`, just listen `focus` on `radio-option` and proxy `disabled` state

common.blocks/base-control/base-control.js
```diff
@@ -0,0 +1,108 @@
...
+        this.elem('control').blur();
+    }
+}, /** @lends base-control */{
+    live : function() {
+        this
+            .liveBindTo('control', 'focusin', this.prototype._onFocus)
+            .liveBindTo('control', 'focusout', this.prototype._onBlur);
+
+        var focused = dom.getFocused();
+        if(focused.hasClass(this.buildClass('control'))) {
```
narqo, line 96: @dfilatov I don't understand how to properly write specs for this case. 

In all of our specs we're running `BEMDOM.int` for some context inside beforeEach hooks. But in this case base-controls's `live` method runs before any suit-case runs. So if I would do something like this (like we did in [menu](https://github.com/bem/bem-components/blob/v2/common.blocks/menu/menu.spec.js#L73)):

``` js
it(..., function() {
  var domElem = $(BEMHTML.apply(bemjson)).appendTo('body');
  domElem.find('.' + BaseControl.buildClass('control')).focus();

  BEMDOM.init(domElem);
})
```

block would be inited with `focusin` listener from live section, not because of it's DOM-focus.

dfilatov, line 96: I think you can try to use some way like here, can't you?
https://github.com/bem/bem-components/blob/v2/common.blocks/input/input.spec.js#L42

narqo, line 96: Unfortunately no. In this example, `input` would be inited after [focus()](https://github.com/bem/bem-components/blob/v2/common.blocks/input/input.spec.js#L44) method has been called. Not because of `BEMDOM.init` :(

dfilatov, line 96: I don't understand. live-section of block declaration will be invoked when the first such block is found during `BEMDOM.init`, not during `BEMDOM.decl`.

narqo, line 96: Yep, but first `BEMDOM.init` is called during first `beforeEach`. So every next block appended to `body` would be inited with `liveBindTo`-handlers, isn't it?

common.blocks/menu/menu.js
```diff
@@ -29,30 +22,26 @@ provide(BEMDOM.decl(this.name, /** @lends menu.prototype */{
...
-                this._focused && this._blur();
                 this._hoveredItem && this._hoveredItem.delMod('hovered');
+                this.__base.apply(this, arguments);
             }
         },
 
         'disabled' : function(modName, modVal) {
-            this.domElem.attr('tabindex', modVal? -1 : 0);
+            this.__base.apply(this, arguments);
+            this.elem('control').attr('tabindex', modVal? -1 : 0);
```
dfilatov: `removeAttr('tabindex')` is better in that case.

narqo:
@dfilatov :up:

common.blocks/input/input.js
```diff
@@ -1,53 +1,36 @@
...
      * Returns control value
      * @returns {String}
+     * @override
      */
     getVal : function() {
         return this._val;
     },
 
+
+
```
dfilatov: I think you can leave one empty line instead of three )
Btw, you can add `disallowMultipleLineBreaks` option in jscs config to prevent that in the further.

narqo: Oh, shit :( Fixed

> Btw, you can add disallowMultipleLineBreaks option in jscs config to prevent that in the further

done in `v2`'s HEAD

dfilatov:
lgtm

PR #383, opened 2014-04-11
Infrastructure improvements

tadatuta:
- used `roole` tech from `bem-tools` (so it made it possible to remove `q` and `borschik` from `package.json`)
- stoped extending `bem make` with `bem-core` nodes (everything useful seems to be already included in `bem-tools`)
- updated a few npm dependencies

// cc @narqo


package.json
```diff
@@ -29,12 +27,14 @@
...
     "git-hooks": "0.0.6",
     "svgo": "0.4.1",
-    "bower-npm-install": "~0.5.2",
+    "bower-npm-install": "~0.5.4",
     "istanbul": "^0.2.7"
   },
   "optionalDependencies": {},
   "private": true,
   "scripts": {
+    "postinstall": "bower-npm-install --non-interactive",
```
narqo: No way! https://github.com/narqo/bem-game-go/issues/7

package.json
```diff
@@ -15,13 +15,11 @@
   },
   "dependencies": {
     "bem-environ": "~1.4.0",
-    "q": "~0.9.7",
-    "roole": "~0.9.0",
-    "borschik": "~0.4.0"
+    "roole": "~0.9.0"
```
narqo: Do we still need roole itself?

tadatuta: yes, techs do not provide npm-dependencies

package.json
tadatuta: removed

package.json
narqo: But how could you guarantee that the `roole` module that I have in my package.json will be compatible with your (bem-tools') tech? What would happened if roole 1.0 would have backward incompatible API?

tadatuta: it's a common problem for all of our preprocessor techs but until you specify `bem@0.8.0` and `roole@0.9.0` in project's dependencies it's kinda guaranteed to work fine. so I don't think it's rather big problem to add preprocessors into `peerDependencies` of `bem-tools`. and as soon as we have quite a few of such techs it's possible to release updates in almost no time.

narqo: Don't you think that this provides a very strange user experience (even for us)? 

How would you explain to users the need to add "proper" version of `roole` to there dependencies? Even if it is included as a `dependencies` in bem-components. Shouldn't it be moved to `devDependencies` in this case?

package.json
```diff
@@ -15,13 +15,11 @@
   },
   "dependencies": {
     "bem-environ": "~1.4.0",
```
narqo: I have a filling that we don't need `bem-environ` in bem-component's `dependencies` as well. Someone really should  check this.

For now we have a problem, that bower-npm-install installs npm-dependencies for every dependent-library we have. So a lot of npm-modules are installed several times. This crap, increase time spent for dependencies installation. And this is very sad if you try to work in places with slow WiFi, like cafe or hotel.

package.json
tadatuta: It's ok for me to move it to `devDependencies`. I agree it's rather ugly but I don't know better variant :(

package.json
tadatuta: now it's used in `make.js` and `level.js` to get paths to libraries and for `getTechResolver` method. so it looks useful enough.

narqo: The question is could we move to `devDependencies`? Initially it was make for the goal of proper tech-path resolving in nested dependencies. So it couldn't be used from dev-deps. But since we don't have any techs outside of bem-core, may be we could finally move it to dev?

tadatuta:
@narqo I tried to build `project-stub` without installing `bem-environ` inside of `bem-components` and it failed with

```
Error: Tech module with path '/Users/tadatuta/Sites/pr-stub-core/libs/bem-components/libs/bem-core/.bem/techs/browser.js.js' not found on require search paths
    at exports.Level.INHERIT.resolveTechPath (/Users/tadatuta/Sites/pr-stub-core/node_modules/bem/lib/level.js:280:19)
    at exports.Level.INHERIT.resolveTech (/Users/tadatuta/Sites/pr-stub-core/node_modules/bem/lib/level.js:228:25)
    at exports.Level.INHERIT.resolveTechName (/Users/tadatuta/Sites/pr-stub-core/node_modules/bem/lib/level.js:244:47)
    at exports.Level.INHERIT.resolveTech (/Users/tadatuta/Sites/pr-stub-core/node_modules/bem/lib/level.js:231:25)
    at exports.getTechClass (/Users/tadatuta/Sites/pr-stub-core/node_modules/bem/lib/tech/index.js:42:44)
    at exports.createTech (/Users/tadatuta/Sites/pr-stub-core/node_modules/bem/lib/tech/index.js:81:21)
    at Object.pathToName (/Users/tadatuta/Sites/pr-stub-core/node_modules/bem/lib/context.js:26:25)
    at Array.map (native)
    at Object.exports.Context.INHERIT.initOptsTechs (/Users/tadatuta/Sites/pr-stub-core/node_modules/bem/lib/context.js:34:45)
    at Object.exports.Context.INHERIT.__constructor (/Users/tadatuta/Sites/pr-stub-core/node_modules/bem/lib/context.js:16:27)
```

so I think we can merge this PR as is (I just rebased it)?

package.json
```diff
@@ -14,27 +14,26 @@
...
     "git-hooks": "0.0.6",
     "svgo": "0.4.1",
-    "bower-npm-install": "~0.5.2",
+    "bower-npm-install": "~0.5.4",
     "istanbul": "^0.2.7"
   },
   "optionalDependencies": {},
   "private": true,
   "scripts": {
+    "deps": "bower-npm-install --non-interactive",
```
narqo, line 36: Don't you think, that running `bower-npm-install` is much clearer than `npm run deps`? Why do we need this?

narqo:
I still don't like the necessity to have `roole` dependency in all dependent libraries just because `bem-tools` is suxx. But if everyone happy with this, ship it

package.json
tadatuta, line 36: that's quite handy for users without `$PATH` hack and is a kind of abstraction which can be used for some other deps in the future.

tadatuta:
as `roole` was there anyway let's think about improvement later ;)

Issue #388, opened 2014-04-14
input: cursor position bug


sipayRT:
while changing request in the middle of the line the cursor returns to the end of the line

persidskiy:
What do you mean? Could you provide a screenshot?

PR #393, opened 2014-04-15
input: fix cursor's jump to end of input when user edit text at the middle of input

persidskiy:
Solve #388 

/cc @dfilatov @narqo

### DIFF
--- a/common.blocks/input/input.js
+++ b/common.blocks/input/input.js
@@ -59,7 +59,10 @@ provide(BEMDOM.decl(this.name, {
 
         if(this._val !== val) {
             this._val = val;
-            this.elem('control').val(val);
+
+            var control = this.elem('control');
+            control.val() !== val && control.val(val);
+
             this.emit('change', data);
         }
 


common.blocks/input/input.js
```diff
@@ -55,11 +55,17 @@ provide(BEMDOM.decl(this.name, {
...
     setVal : function(val, data) {
+        var control = this.elem('control');
+
         val = String(val);
 
         if(this._val !== val) {
             this._val = val;
-            this.elem('control').val(val);
+
+            if(control.val() !== val) {
```
dfilatov: We use `&&` for short conditions.

``` js
control.val() !== val && control.val(val);
```

persidskiy: Ok

common.blocks/input/input.js
```diff
@@ -55,11 +55,17 @@ provide(BEMDOM.decl(this.name, {
      * @returns {this}
      */
     setVal : function(val, data) {
+        var control = this.elem('control');
```
dfilatov: Seems that variable `control` is used only inside the next `if`. It should be declared nearer to place of using.

persidskiy: I had thought same, but decided to push all `var` declarations to top of the function.  :)

persidskiy:
:up:

common.blocks/input/input.js
```diff
@@ -58,8 +58,12 @@ provide(BEMDOM.decl(this.name, {
         val = String(val);
 
         if(this._val !== val) {
+            var control = this.elem('control');
```
dfilatov: And the final touch -- move this declaration to place before its first using.

persidskiy:
done

PR #394, opened 2014-04-16
icon: interface for inline image url setting (close #392)

### LINKED ISSUE #392: icon: Provide interface for inline image url setting

mishanga:


mishanga:
@veged merge it if okay.

PR #398, opened 2014-04-16
button: Move keyboard related stuff into common.blocks (close #359)

### LINKED ISSUE #359: button: Move keyboard related stuff into common.blocks

dfilatov:
cc @narqo @veged

### DIFF
--- a/common.blocks/button/button.js
+++ b/common.blocks/button/button.js
@@ -1,7 +1,10 @@
 modules.define(
     'button',
-    ['i-bem__dom', 'jquery', 'dom'],
-    function(provide, BEMDOM, $, dom) {
+    ['i-bem__dom', 'jquery', 'dom', 'functions'],
+    function(provide, BEMDOM, $, dom, functions) {
+
+var KEY_CODE_SPACE = 32,
+    KEY_CODE_ENTER = 13;
 
 provide(BEMDOM.decl(this.name, {
     beforeSetMod : {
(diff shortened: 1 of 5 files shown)


common.blocks/button/button.js
```diff
@@ -1,7 +1,10 @@
 modules.define(
     'button',
-    ['i-bem__dom', 'jquery', 'dom'],
-    function(provide, BEMDOM, $, dom) {
+    ['i-bem__dom', 'jquery', 'dom', 'functions'],
+    function(provide, BEMDOM, $, dom, functions) {
+
+var KEY_CODE_SPACE = 32,
+    KEY_CODE_ENTER = 13;
```
narqo, line 7: as we don't not fix our bem-core's version in bem-components, maybe we could use `keyboard__codes` from current v2's `HEAD`?

common.blocks/button/button.spec.js
```diff
@@ -61,21 +61,37 @@ describe('button', function() {
...
+            button.hasMod('pressed').should.be.true;
+        });
+
+        it('should not be pressed on pointerrelease', function() {
             button.domElem.trigger('pointerrelease');
             button.hasMod('pressed').should.be.false;
         });
 
-        it('should not set "hovered" mod if it has "disabled" mod', function() {
+        it('should not be pressed if it is disabled', function() {
```
narqo: Guess you could say "(it) should not be pressed if disabled" :)

narqo:
looks good

common.blocks/button/button.js
dfilatov, line 7: I don't understand how bower resolve that case. I've tried:

```
rm -rf libs
bower-npm-install
```

I've got:

```
bower bem-core#v2             validate 2.1.0 against git://github.com/bem/bem-core.git#v2
```

WTF? Why has bower installed 2.1.0?

common.blocks/button/button.spec.js
dfilatov: fixed

common.blocks/button/button.js
narqo, line 7: Magic :(

PR #409, opened 2014-04-20
menu: Do not add tabindex attribute if disabled mod was set before init

### LINKED ISSUE #407: menu: Do not add "tabindex" attribute if "disabled" mod was set before init

narqo:
closes #407

### DIFF
--- a/common.blocks/menu/menu.bemhtml
+++ b/common.blocks/menu/menu.bemhtml
@@ -1,13 +1,17 @@
 block('menu')(
-    attrs()({ tabindex : 0, role : 'menu' }),
-    js()(true),
     def()(function() {
         applyNext({ _menuTheme : this.mods.theme });
         delete this._menuTheme;
-    })
+    }),
+    attrs()(function() {
+        var attrs = { role : 'menu' };
+        this.mods.disabled || (attrs.tabindex = 0);
+        return attrs;
+    }),
+    js()(true)
 )
 
 block('menu-item').match(this._menuTheme).def()(function() {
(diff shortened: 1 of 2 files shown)


Guria:
What was a reason to made this change? It causes `menu` block unable to work as expected if it appears disabled initially. Control will not be able to get `tabIndex` attribute on mod `disabled` removed anymore. So it will be treated as unfocusable.
//cc @tadatuta 
I ask to revert this change. As `control` descendant `menu` block will store tabIndex in instance as `this._tabIndex` and remove attribute on domNode once block is inited.

tadatuta:
cc @narqo

qfox:
@Guria :+1:

narqo:
We can't just revert these changes. The problem is that `menu` should not be "focusable" while it is disabled. But: 
- at the time such initially disabled block would get DOM-focus, it will be initialized;
- we would get indeterminate state: block is in the focus state, but there is no `_focused` modifier.

These problems are fixed in native "focusable" elements, with `disable` attribute. Could you open a separate issue about initially disabled `menu` – we would think about the right solution.

Guria:
@narqo Ok, i will open an issue.
Is this problem just about menu, or other blocks that use tabIndex also affected? See #1190 for reference.
Also, IMO, a short moment when block in indeterminate state is better then block that doesn't work.

narqo:
> Is this problem just about menu, or other blocks that use tabIndex also affected?

I think, currently the problem is only with `menu`. But will consider this case in other blocks with #1206

PR #415, opened 2014-04-22
Spin in button doesn't look good (fix #414)

sipayRT:

PR #416, opened 2014-04-22
add button__icon element

### LINKED ISSUE #414: Spin in button doesn't look good
``` js
{
    block: 'button',
    mods: { theme: 'normal', size: 's' },
    content: [
        { block: 'spin', mods : { theme : 'normal', size : 'xs', progress : true } },
        { elem: 'text', content: 'Loading...' }
    ]
}
```

In result we have this:
http://yadi.sk/d/Zc4JPLvrMvAzj

persidskiy:
Close #414


mishanga:
I just got rid of it in #385 :)

mishanga:
Instead of `button__icon` we can make `icon_spin`, mixing `icon` and `spin` blocks.

persidskiy:
@mishanga I thought about creating `icon_spin` that was having animation to spin any icon :)
It this case, we can implement `spin` block as regular svg-icon to keep simplicity of `icon`.

Btw, i think, that we should keep the `button__icon` element, and mix it to icon for reduce cascading in button styles. Keep it in BEM-way :)

mishanga:
@Coltspb take a look on https://github.com/bem/bem-components/pull/415/files

PR #417, opened 2014-04-22
Use keyboard__codes

### LINKED ISSUE #408: Use bem-core's keyboard__codes instead of private constants
Since bem-core 2.2.0 we have [keyboard__codes](https://github.com/bem/bem-core/blob/v2.2.0/common.blocks/keyboard/__codes/keyboard__codes.js) module to use where private constants are used now

narqo:
closes #408

cc @dfilatov

### DIFF
--- a/common.blocks/button/button.deps.js
+++ b/common.blocks/button/button.deps.js
@@ -6,6 +6,7 @@
             mods : { focused : 'yes', hovered : 'yes', disabled : 'yes', pressed : 'yes' }
         },
         { block : 'jquery', elem : 'event', mods : { type : 'pointer' } },
+        { block : 'keyboard', elem : 'codes' },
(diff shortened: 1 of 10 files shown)


dfilatov:
Spec "menu hover on items should react on up/down keys when focused" has failed.

narqo:
Crap, writing code during fly, without dependencies installed was not a good idea :) Fixed

dfilatov:
But PR is still failed :)

persidskiy:
It was a good idea to separate key codes into module. Looks really better now :+1:

narqo:
> But PR is still failed :)

This time it was "Mocha timeout" bug :(

dfilatov:
LGTM

PR #419, opened 2014-04-22
add getText()/setText() methods to button

### LINKED ISSUE #418: Add setText() and getText() methods to button.
It is useful to have such methods directly in `button` block.

persidskiy:
Close #418

@dfilatov @narqo

### DIFF
--- a/common.blocks/button/button.js
+++ b/common.blocks/button/button.js
@@ -54,6 +54,25 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : BaseControl }, /** @lends b
         }
     },
 
+    /**
+     * Returns text of the button
+     * @returns {String}
+     */
+    getText : function() {
+        return this.elem('text').text();
+    },
+
+    /**
+     * Sets text to the button
+     * @param {String} text
+     * @returns {this}
+     */
+    setText : function(text) {
+        this.elem('text').text(text);
+
+        return this;
+    },
+
     _onUnload : function() {
         this.delMod('focused');
     },
--- a/common.blocks/button/button.spec.js
+++ b/common.blocks/button/button.spec.js
@@ -7,7 +7,7 @@ describe('button', function() {
     var button;
 
     beforeEach(function() {
-        button = BEMDOM.init($(BEMHTML.apply({ block : 'button' })).appendTo('body'))
+        button = BEMDOM.init($(BEMHTML.apply({ block : 'button', text : 'foo' })).appendTo('body'))
             .bem('button');
     });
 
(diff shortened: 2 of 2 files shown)


common.blocks/button/button.spec.js
```diff
@@ -140,6 +145,22 @@ describe('button', function() {
...
+            button.getText(1).should.be.equal('world');
+        });
+
+        it('should update text in the text element by index', function() {
+            button.setText('hi');
+            button.elem('text').eq(0).text().should.be.equal('hi');
+
+            button.setText('universe', 1);
+            button.elem('text').eq(1).text().should.be.equal('universe');
+        });
```
persidskiy, line 152: Should I add tests for html-escaping here?

dfilatov:
Why do we need to have multiple text element?

common.blocks/button/button.js
```diff
@@ -57,6 +57,31 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : BaseControl }, /** @lends b
...
+     * @returns {String}
+     */
+    getText : function(index) {
+        return this._getTextByIndex(index).text();
+    },
+
+    /**
+     * Sets text to the `text` element
+     * @param {String} text
+     * @param {Number} [index=0] index of the text element
```
apostololeg: слишком уж низкоуровнево, как для внешнего API

persidskiy:
Hmm... you asked a good question. I saw the [examples](https://github.com/bem/bem-components/blob/v2/common.blocks/button/button.tests/simple.bemjson.js#L173-179) of using multiple `text` elems and decided it can be useful, but I could not find any real cases of using it.

I will get rid of these indexes :)

common.blocks/button/button.js
persidskiy: Да, после поиска примеров использования, кажется, что уже и не надо :)

persidskiy:
:up:

dfilatov:
LGTM

common.blocks/button/button.js
```diff
@@ -54,6 +54,25 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : BaseControl }, /** @lends b
         }
     },
 
+    /**
+     * Returns text form the text element
```
narqo: "form" (sic)?

common.blocks/button/button.spec.js
```diff
@@ -140,6 +140,17 @@ describe('button', function() {
             spy.should.not.have.been.called;
         });
     });
+
+    describe('setText/getText', function() {
+        it('should return text from the text element by index', function() {
```
narqo: guess, "by index" part is obsolete

common.blocks/button/button.js
dfilatov: I think we should avoid using inner details in jsdoc.

persidskiy:
:up:

PR #423, opened 2014-04-28
Some improvements and fixes for IE

dfilatov:
cc @narqo @veged

### DIFF
--- a/common.blocks/popup/popup.js
+++ b/common.blocks/popup/popup.js
@@ -4,8 +4,8 @@
 
 modules.define(
     'popup',
-    ['i-bem__dom', 'jquery', 'dom', 'functions__throttle'],
-    function(provide, BEMDOM, $, dom, throttle) {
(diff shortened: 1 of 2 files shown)


common.blocks/popup/popup.js
```diff
@@ -15,9 +15,10 @@ var VIEWPORT_ACCURACY_FACTOR = 0.99,
         'left-top', 'left-center', 'left-bottom'
     ],
     BASE_ZINDEX = 10000,
-    CHECK_OWNER_THROTTLING_INTERVAL = 100,
+    UPDATE_TARGET_VISIBILITY_THROTTLING_INTERVAL = 100,
```
narqo, line 18: Guess we should rename this in `popup.spec` as well?

narqo:
:+1:

common.blocks/popup/popup.js
dfilatov, line 18: fixed

PR #426, opened 2014-04-28
Don't use live init for already focused buttons (close #424)

### LINKED ISSUE #424: button: correct '_focused' modifier behavior
Two buttons at once can be focused on a page. If a '_focused_true' modifier is set for one button and you click the another one (in range of one theme), this button becomes focused too. 
Checked for simple and normal themes
Link to screenshot: http://yadi.sk/d/AdD6kUAUNVF2Q

dfilatov:
cc @narqo

### DIFF
--- a/common.blocks/button/_focused/button_focused.bh.js
+++ b/common.blocks/button/_focused/button_focused.bh.js
@@ -1,0 +1,9 @@
+module.exports = function(bh) {
+
+    bh.match('button_focused', function(ctx, json) {
+        ctx
+            .applyBase()
+            .js(ctx.extend(json.js, { live : false }), true);
+    });
+
+};
--- a/common.blocks/button/_focused/button_focused.bemhtml
+++ b/common.blocks/button/_focused/button_focused.bemhtml
@@ -1,0 +1,3 @@
+block('button').mod('focused', true).js()(function() {
+    return this.extend(applyNext(), { live : false });
+})


common.blocks/button/_focused/button_focused.bh.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function(bh) {
+
+    bh.match('button_focused', function(ctx, json) {
+        ctx
+            .applyBase()
+            .js(ctx.extend(json.js, { live : false }), true);
```
dfilatov, line 6: @mishanga take a look please. I don't know right way to add param to js.

mishanga, line 6: Better to use `ctx.js()` instead of `json.js`. Looks like it's not equal.

dfilatov, line 6: What's the difference?

narqo:
:+1:

common.blocks/button/_focused/button_focused.bh.js
tadatuta, line 6: @mishanga PR is waiting for your answer

mishanga, line 6: ~~Oops :) Everything is okay, no difference: https://github.com/enb-make/bh/blob/master/lib/bh.js#L459~~

UPD:
@dfilatov there is the difference between `json` and `ctx.json()`, but no difference between `json.js` and `ctx.js()`:

``` javascript
bh.match('button', function(ctx) {
    ctx.js({ foo: 1 });
});
bh.match('button', function(ctx, json) {
    console.log(1, json === ctx.json());
    console.log(1, ctx.js(), json.js);
    ctx.applyBase();
    console.log(2, json === ctx.json());
    console.log(2, ctx.js(), json.js);
});
bh.apply({ block: 'button' })
```

Result:

```
1 true
1 undefined undefined
2 false
2 { foo: 1 } { foo: 1 }
```

PR #438, opened 2014-04-30
clearer code for radio

dfilatov:
cc @narqo @veged

### DIFF
--- a/common.blocks/radio-group/radio-group.js
+++ b/common.blocks/radio-group/radio-group.js
@@ -51,8 +51,7 @@ provide(BEMDOM.decl(this.name, /** @lends radio-group.prototype */{
                     i = 0, radio;
 
                 while(radio = radios[i++]) {
-                    if(!radio.hasMod('disabled')) {
-                        radio.setMod('focused');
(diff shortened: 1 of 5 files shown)


narqo:
Cool, but maybe in addition to your changes we could do a little bit more: 

1) in https://github.com/bem/bem-components/blob/radio-imp/common.blocks/radio-group/radio-group.js#L53-L58 – as I remember, `radio` augments `base-control`, so it should be impossible to set `_focused` if one has `_disabled`
2) https://github.com/bem/bem-components/blob/radio-imp/common.blocks/radio-group/radio-group.js#L124 – guess this variable should be called `radios`
3) Why do we use `forEach` here, https://github.com/bem/bem-components/blob/radio-imp/common.blocks/radio-group/radio-group.js#L139, but `while` in other cases

dfilatov:
@narqo 
1) There is another reason to check for `disabled`: I need to find first radio that's able to be focused. 
I've remade a little bit that code and I've added test, take a look.
2) Fixed
3) Because I need to break cycle depending on condition, but using of `some()` for that purpose is unacceptable for me.

narqo:
Oh, now I see, tnx. Looks clear

PR #464, opened 2014-05-07
link: Make identical bh-template as bemhtml

### LINKED ISSUE #443: link: Make identical bh-template as bemhtml

sipayRT:
Close #443 

@mishanga

### DIFF
--- a/common.blocks/link/link.bh.js
+++ b/common.blocks/link/link.bh.js
@@ -14,7 +14,7 @@ module.exports = function(bh) {
 
         typeOfUrl !== 'undefined' && (attrs.href = typeOfUrl === 'string'?
             url :
-            url); // TODO: реализовать возможность отдавать bemjson в url
+            bh.apply(url));
 
         typeof attrs.href === 'undefined' &&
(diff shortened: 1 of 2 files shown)


common.blocks/link/link.bh.js
```diff
@@ -14,7 +14,7 @@ module.exports = function(bh) {
 
         typeOfUrl !== 'undefined' && (attrs.href = typeOfUrl === 'string'?
             url :
-            url); // TODO: реализовать возможность отдавать bemjson в url
+            bh.apply(url));
```
mishanga, line 17: `ctx.apply(url);`

mishanga:
В остальном ок.

common.blocks/link/link.bh.js
sipayRT, line 17: at this case `ctx.apply()` returned `[object Object]` instead of `/action`.

mishanga, line 17: Oops! Yes, of course, you are right.

PR #492, opened 2014-05-16
menu-item block description 

Murdalay:


tadatuta:
влил сосквошенный PR https://github.com/bem/bem-components/pull/493

PR #525, opened 2014-05-28
Border disappears in pseudobutton's disabled state

gela-d:
/cc @mishanga

### DIFF
--- a/design/common.blocks/button/_theme/button_theme_normal.roo
+++ b/design/common.blocks/button/_theme/button_theme_normal.roo
@@ -137,7 +137,7 @@
     {
         overflow: hidden;
 
-        background: 0 !important;
+        background: 0;
 
         &:before
         {


gela-d:
@mishanga Mishaaaangaaa :)

mishanga:
21_podborka_31

design/common.blocks/button/_theme/button_theme_normal.roo
```diff
@@ -178,7 +178,7 @@
         color: rgba(0, 0, 0, 0.3);
         background: rgba(0, 0, 0, 0.08);
 
-        &:before
+        &:not(.button_pseudo):before
```
mishanga: This is wrong way. You should just remove `important` from here:
https://github.com/bem/bem-components/blob/126deac621f3174566d63530a9d827b97ac21086/design/common.blocks/button/_theme/button_theme_normal.roo#L132

gela-d:
@mishanga mimimi :)
Done.

PR #539, opened 2014-05-29
checkbox(-group), radio(-group): css, refactoring

### LINKED ISSUE #508: checkbox-group: make css

mishanga:
close #508 
By default radio and checkboxes inside their groups are divided by line break.
Line view (with proper margins) realized in `type_line` modifier.

### DIFF
--- a/common.blocks/button/button.tests/simple.bemjson.js
+++ b/common.blocks/button/button.tests/simple.bemjson.js
@@ -286,55 +286,29 @@
         ] },
 
         { tag : 'p', content : [
-            { tag : 'span', content : {
+            {
                 block : 'button',
                 mods : { theme : 'normal', size : 'm', togglable : 'check' },
                 text : 'check'
-            } },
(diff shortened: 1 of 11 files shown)


mishanga:
@dfilatov @narqo

persidskiy:
@mishanga `button_type_check` doesn't look good. 
2014-05-29 19-55-08 bem-components button

BTW, may be we need to invent common approach for inlining different controls. For instance button + button, input+button, input+input, radio-groups, check-groups. I guess our designers like these combinations. 
input


I thought about certain block (`control-group`, for example), which has elements `__button` and `__input`, which can be mixed to `button` or `input` blocks to make them aligned to each other.

What do you think about this idea? ;)

mishanga:
@Coltspb thanks for the note!
And as for `control-group` — I am totally agree with you. Yesterday we talked about it with @dfilatov. Here is the issue: #540

PR #605, opened 2014-06-11
Add test for 'input' block

### LINKED ISSUE #583: input: Check coverage tests

eGavr:
closes #583

### DIFF
--- a/common.blocks/input/input.spec.js
+++ b/common.blocks/input/input.spec.js
@@ -15,6 +15,12 @@ describe('input', function() {
     });
 
     describe('val', function() {
+        it('should return new value when value changed', function() {
+            input
+                .setVal('blabla')
+                .getVal().should.be.equal('blabla');
+        });
+
         it('should update control elem value when value changed', function() {
             input.setVal('blabla');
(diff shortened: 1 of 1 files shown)


common.blocks/input/input.spec.js
```diff
@@ -15,6 +15,11 @@ describe('input', function() {
     });
 
     describe('val', function() {
+        it('checking of \'getter\' and \'setter\'', function() {
```
tadatuta: it's better to use `should [do something]` form. e.g. `should return new value when value changed`

common.blocks/input/input.spec.js
```diff
@@ -15,6 +15,11 @@ describe('input', function() {
...
+            input.setVal('blabla');
```
tadatuta: I see it's not used in other tests but we can chain code like this:

``` js
input
    .setVal('blabla')
    .getVal().should.be.equal('blabla');
```

gela-d:
@tadatuta can we merge?

tadatuta:
LGTM

PR #638, opened 2014-06-13
input_disabled: In disabled state cursor in Fx should be type='text'

### LINKED ISSUE #636: input_disabled: In disabled state cursor in Fx should be type='text'
I can't screen my screen with incorrect cursor, but trust me, there is a bug :)

gela-d:
Close #636


gela-d:
@mishanga ? :)

PR #639, opened 2014-06-13
input__label: Doesn't have any opacity in disabled state

### LINKED ISSUE #637: input__label: Doesn't have any opacity in disabled state

gela-d:
Close #637


gela-d:
@mishanga :)

PR #674, opened 2014-06-18
menu: example for scrolled content (close #555)

### LINKED ISSUE #555: menu: css for scrolled content
In some browsers scrollbar has width that causes problems with `overflow: ellipsis`

mishanga:
close #555


mishanga:
@dfilatov merge it if ok.

dfilatov:
@mishanga It seems we don't need to do that?

mishanga:
Yep, we don't.

PR #680, opened 2014-06-18
dropdown: Write Gemini tests 

DimaChervan:


gela-d:
new PR - https://github.com/bem/bem-components/pull/701

PR #695, opened 2014-06-23
select: make it work

dfilatov:


common.blocks/select/_type/select_type_radio.js
```diff
@@ -11,7 +11,8 @@ modules.define('select', function(provide, Select) {
  */
 provide(Select.decl({ modName : 'type', modVal : 'radio' }, /** @lends select.prototype */{
     _updateControl : function() {
-        this.elem('control').val(JSON.stringify(this.getVal()));
+        var val = this.getVal();
+        this.elem('control').val(typeof val === 'string'? val : JSON.stringify(val));
```
narqo: Maybe we could check `typeof val !== 'object'? val : JSON.strinigify(val)`? Guess jQuery would do the rest.

common.blocks/select/select.js
```diff
@@ -230,7 +232,7 @@ provide(BEMDOM.decl(this.name, /** @lends select.prototype */{
             'type="hidden" ' +
             'name="' + name + '" ' +
             'class="' + this.buildClass('control') + '" ' +
-            'value="' + escape.attr(JSON.stringify(val)) + '"/>';
+            'value="' + escape.attr(typeof val === 'object'? JSON.stringify(val) : val) + '"/>';
```
narqo, line 235: :)

narqo:
looks complicated but right :) :+1:

common.blocks/select/_type/select_type_radio.js
dfilatov: fixed

PR #730, opened 2014-06-27
menu: Pass disabled mod to every nested menu-item

### LINKED ISSUE #352: menu: pass disabled mod to every nested menu-item

sipayRT:
Close #352

### DIFF
--- a/common.blocks/menu/menu.bh.js
+++ b/common.blocks/menu/menu.bh.js
@@ -1,9 +1,14 @@
 module.exports = function(bh) {
 
     bh.match('menu', function(ctx) {
+        var menuMods = {
+            theme : ctx.mod('theme'),
+            disabled : ctx.mod('disabled')
+        };
+
         ctx
             .js(true)
-            .tParam('_menuTheme', ctx.mod('theme'))
+            .tParam('_menuMods', menuMods)
             .mix({ elem : 'control' });
 
         var attrs = { role : 'menu' };
@@ -12,7 +17,12 @@ module.exports = function(bh) {
     });
 
     bh.match('menu-item', function(ctx) {
-        ctx.mod('theme', ctx.tParam('_menuTheme'));
+        var menuMods = ctx.tParam('_menuMods');
+
+        ctx.mods({
+            theme : menuMods.theme,
+            disabled : menuMods.disabled
+        });
     });
 
 };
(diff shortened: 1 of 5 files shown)


dfilatov:
@sipayRT you should do that in js as well

sipayRT:
@dfilatov check, plz

common.blocks/menu/menu.js
```diff
@@ -37,6 +37,14 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : BaseControl }, /** @lends m
                     .__base.apply(this, arguments);
                 this._hoveredItem && this._hoveredItem.delMod('hovered');
             }
+        },
+
+        'disabled' : {
```
dfilatov: You can reduce nesting:

``` js
'disabled' : function(modName, modVal) {
    this.getItems().forEach(function(menuItem){
        menuItem.setMod(modName, modVal);
    });
}
```

common.blocks/menu/menu.bh.js
```diff
@@ -4,6 +4,7 @@ module.exports = function(bh) {
         ctx
             .js(true)
             .tParam('_menuTheme', ctx.mod('theme'))
+            .tParam('_menuDisabled', ctx.mod('disabled'))
```
dfilatov: Might be we'll pass `_menuMods` instead of passing each modifier separately?

sipayRT:
@dfilatov fixed and rebased

dfilatov:
@sipayRT we need to add specs for js behaviour

sipayRT:
@dfilatov thanks for reminding me - there was an error in BEMHTML template

common.blocks/menu/menu.spec.js
```diff
@@ -49,6 +49,13 @@ describe('menu', function() {
             menu.delMod('disabled');
             control.attr('tabindex').should.be.equal('0');
         });
+
+        it('should pass disabled mod to menu items', function() {
+            menu.setMod('disabled');
+            menuItems.forEach(function(menuItem) {
+                menuItem.hasMod('disabled').should.be.true;
+            });
+        });
```
dfilatov, line 62: Add a test case for the opposite situation -- all `menu-item` should be enabled while `menu` is being enabled

dfilatov, line 62: @sipayRT ping

sipayRT:
@dfilatov sorry for the delay - we had holidays. Specs was added

dfilatov:
:shipit:

PR #737, opened 2014-06-30
Use bem-core with fixed pointer events

narqo:
See bem/bem-core#567


narqo:
Maybe we could ask our testing-guys to check everything in all supported devices?

\cc @gela-d @mishanga

narqo:
Gemini-tests fail, where should I look to see diff images?

PR #805, opened 2014-07-09
button: Fix BH template

### LINKED ISSUE #804: button: Fix BH template

sipayRT:
Close #804 
/cc @dfilatov 

If we do not add any `type` mods, so in HTML it will be `type=""`. Expected that `type` will not be added


tadatuta:
lgtm

PR #810, opened 2014-07-10
popup_theme_normal: Completely hide popup until one is initialized

### LINKED ISSUE #809: popup_theme_normal: Completely hide popup until one is initialized
According to specs, elements with `visibility: hidden` property set are still occupy space in the layout, so the could affect initial page Layout time. 

At the same time we could fix that without any cost, with hiding `popup`s until they become initialized (e.g. got `popup_js_inited` CSS class).

See "[Render-tree construction, Layout, and Paint](https://developers.google.com/web/fundamentals/performance/critical-rendering-path/render-tree-construction)" article (for example) for inspiration ;)

narqo:
closes #809 

\cc @mishanga @tadatuta @dfilatov

### DIFF
--- a/design/common.blocks/popup/_theme/popup_theme_normal.styl
+++ b/design/common.blocks/popup/_theme/popup_theme_normal.styl
@@ -1,5 +1,6 @@
 .popup_theme_normal
 {
+    display: none;
     visibility: hidden;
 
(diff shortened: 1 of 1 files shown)


mishanga:
ok

PR #847, opened 2014-07-15
block 'select_mode_radio-check' js fix: apply JSON.stringify only for object-values

aliosv:
/cc @dfilatov


tadatuta:
@dfilatov, the reason of this PR is that `JSON.stringify()` adds extra quotes to string values.
can we merge it?

dfilatov:
:shipit:

PR #862, opened 2014-07-17
input_theme_normal: add text-overflow: ellipsis

dfilatov:
cc @mishanga @gela-d

### DIFF
--- a/design/common.blocks/input/_theme/input_theme_normal.styl
+++ b/design/common.blocks/input/_theme/input_theme_normal.styl
@@ -83,6 +83,11 @@ ctx = '.input_theme_normal'
         }
     }
 
+    &.input_has-clear .input__control
+    {
+        text-overflow: ellipsis;
+    }
+
     .input__clear
     {
         position: absolute;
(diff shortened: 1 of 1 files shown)


gela-d:
I don't like how `_type_password` looks.
And maybe we add 2px-padding-right to our inputs?

dfilatov:
Might be add `text-overflow` only in case of `_has-clear` and not `_type_password`?

gela-d:
Yes, it would solve both problems

dfilatov:
But it makes me use hellish css-selector (

sipayRT:
why do you think that `:not()` is so ugly?

dfilatov:
What could be more ugly than `.input_theme_normal.input_has-clear:not(.input_type_password) .input__control`?

gela-d:
If you dont' want use `:not`, we can add `ellipsis` for all our `input_has-clear`, but for `.input_type_password` write `text-overflow: clip`?

mishanga:
@gela-d yep, it would be better.

dfilatov:
@gela-d I've updated PR, please, check

mishanga:
:shipit:

PR #886, opened 2014-07-28
select: button text is wrong in case of groups (close #885)

### LINKED ISSUE #885: select: button text is wrong in case of groups

dfilatov:
cc @narqo @mishanga


mishanga:
ok

PR #888, opened 2014-07-28
button: collapsed without given specified text (close #887)

### LINKED ISSUE #887: button: collapsed without specified text

dfilatov:
cc @mishanga

### DIFF
--- a/common.blocks/button/button.bemhtml
+++ b/common.blocks/button/button.bemhtml
@@ -45,7 +45,7 @@ block('button')(
             var ctx = this.ctx,
                 content = [ctx.icon];
             // NOTE: wasn't moved to separate template for optimization
-            ctx.text && content.push({ elem : 'text', content : ctx.text });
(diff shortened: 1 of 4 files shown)


mishanga:
It can be fixed by CSS only:

``` css
.button_theme_normal .button__text:empty:before
{
    content: '\00a0';
}
```

dfilatov:
@mishanga Does it actually depend on `_theme_normal`? It's not about style. Can I put this code to `button__text.styl`?

mishanga:
@dfilatov yep, it really does. There are several ways to fix this bug.

BTW, current bug is caused by this:
https://github.com/bem/bem-components/blob/906aeee5aad9eb2c938934ef205f126cfe6046d5/design/common.blocks/select/_theme/select_theme_normal.styl#L12-L15
AFAIR `.button__text { display: block }` is needed just for select_width_available.

dfilatov:
@mishanga I've remade PR, please check

mishanga:
:+1:

PR #931, opened 2014-08-08
Update and re-enable gemini

SevInf:
Details about config changes are at the first commit message.
Screenshots updated mostly due to a different crop algorithm in newer version of `gemini`.
In case of a failure, html report is published on Github Pages in repo [bem/reports](https://github.com/bem/reports). Link to the report will be printed to the travis log.

[Example report](http://bem.github.io/reports/2437/gemini-report/).

Tests are pretty slow. We can get some boost if we could upload test pages to some external server instead of running server on Travis via SauceConnect proxy.

/cc @arikon @veged @tadatuta


arikon:
@SevInf На мой взгляд всё ок

tadatuta:
@SevInf круто! спасибо!

SevInf:
@veged @dfilatov можете тоже посмотреть? Все ли вас устраивает?

PR #972, opened 2014-08-26
popup: nested popup doesn't consider zIndexGroupLevel of parent popup (closes #970)

### LINKED ISSUE #970: popup: nested popup doesn't consider zIndexGroupLevel of parent popup

persidskiy:
closes #970

### DIFF
--- a/common.blocks/popup/popup.js
+++ b/common.blocks/popup/popup.js
@@ -17,6 +17,8 @@ var VIEWPORT_ACCURACY_FACTOR = 0.99,
     ZINDEX_FACTOR = 1000,
     UPDATE_TARGET_VISIBILITY_THROTTLING_INTERVAL = 100,
 
+    visiblePopupsZIndexes = {},
+
     win = BEMDOM.win,
     undef;
 
@@ -45,13 +47,13 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
     onSetMod : {
         'js' : {
             'inited' : function() {
-                this._parentPopup = null;
+                this._parentPopup = undef;
                 this._owner = null;
                 this._ownerParents = null;
                 this._popupOwner = null;
                 this._pos = null;
                 this._zIndex = null;
-                this._zIndexGroupLevel = undef;
+                this._zIndexGroupLevel = null;
                 this._isAttachedToScope = false;
                 this._isTargetVisible = undef;
                 this._lastDrawingCss = {
@@ -76,13 +78,10 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
 
         'visible' : {
             'true' : function() {
-                this._zIndex = captureZIndex(
-                    typeof this._zIndexGroupLevel === 'undefined'?
-                        this._zIndexGroupLevel = this._calcZIndexGroupLevel() :
-                        this._zIndexGroupLevel);
                 this._owner && (this._ownerParents = this._owner.parents());
 
                 this
+                    ._captureZIndex()
                     .bindTo('pointerpress', this._onPointerPress)
                     ._bindToParentPopup()
                     ._bindToScrollAndResize()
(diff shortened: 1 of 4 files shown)


common.blocks/popup/popup.js
```diff
@@ -444,7 +448,11 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
     },
 
     _calcZIndexGroupLevel : function() {
-        var res = this.params.zIndexGroupLevel;
+        var res = this.params.zIndexGroupLevel,
+            parentPopup = this._getParentPopup();
+
+        parentPopup && (res = Math.max(res, parentPopup._zIndexGroupLevel));
```
dfilatov: I think there is should be `parentPopup._calcZIndexGroupLevel()`

dfilatov: Moreover, it seems that `parentPopup && (res += parentPopup._calcZIndexGroupLevel());`

persidskiy: I'm not sure about addition of zIndexes. For instance, we have three nested popups with `zIndexGroupLevel: 1`. The third popup will calculate his `_zIndexGroupLevel` as `3`. Then we place another popup in block `z-index-group_level_2` somewhere on the page. Now the third popup will overlap last one. Looks like a bug, because in CSS z-indexes the last popup will overlap the third popup.

btw, i think we shouldn't call `_calcZIndexGroupLevel()` on parent block, because `_zIndexGroupLevel` should be calculated  only when popup shows or changes owner. In other cases `_zIndexGroupLevel = null`, which casts to 0 in `Math.max()`.

veged: why it's looks like bug — we have greater zIndexGroup for third popup, so everything is predictable

persidskiy: but it set implictly, so some popup can accidentally overlap another

persidskiy:
I've fixed gemini tests for popup in https://github.com/bem/bem-components/pull/974

common.blocks/popup/popup.js
```diff
@@ -130,24 +128,30 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
...
             }
         } else {
+            var previousOwner = this._owner;
+
             this._pos = { left : left, top : top };
             this._parentPopup = null;
             this._owner = null;
             this._ownerParents = null;
             this._popupOwner = null;
             this._isTargetVisible = true;
```
persidskiy, line 149: btw, why we don't call `.redraw()` here to apply changing position if popup was visible?

common.blocks/popup/popup.js
persidskiy: OK, I'll choose addition because it more flexible way then `Math.max()` =)

common.blocks/popup/popup.js
```diff
@@ -533,6 +540,39 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
         BEMDOM.destruct(this.domElem);
     },
 
+    _getParentPopup : function() {
+        if(this._owner !== null && this._parentPopup === null) {
+            this._parentPopup = this.findBlockOutside(this._owner, this.__self.getName());
```
dfilatov, line 543: It seems that result of method isn't cached in case of parent popup doesn't exist.

persidskiy, line 543: I think we can use `undef` as initial state and `null` as cached absence state.

dfilatov, line 543: Yes, we can

common.blocks/popup/popup.js
```diff
@@ -76,24 +78,20 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
...
             'true' : function() {
-                this._zIndex = captureZIndex(
-                    typeof this._zIndexGroupLevel === 'undefined'?
-                        this._zIndexGroupLevel = this._calcZIndexGroupLevel() :
-                        this._zIndexGroupLevel);
                 this._owner && (this._ownerParents = this._owner.parents());
 
                 this
                     .bindTo('pointerpress', this._onPointerPress)
+                    ._captureZIndex()
```
dfilatov, line 84: Could you move up this line above to group all `bind*` methods?

dfilatov:
in general LGTM

persidskiy:
Guys, could I merge it? =)

Issue #975, opened 2014-08-28
(radio|checkbox)_type_button: won't work in IE9


dfilatov:
IE9 cuts inner input@type=radio|checkbox from DOM

PR #979, opened 2014-08-29
Run tmpl-specs, update bower to 1.3.9

tadatuta:


tadatuta:
@mishanga Please take a look at `select__menu.bh.js`. Now if `select` is disabled, all `menu-item`s are also disabled.

@dfilatov This PR also updates `tmpl-specs` after #975

mishanga:
:+1:

PR #991, opened 2014-09-02
Sync menu theme with menu-items by default

### LINKED ISSUE #861: Sync menu theme with menu-items by default
Now we are to explicitly set `menu-item_theme` even though it's already defined for `menu`. I think we can propagate it if there's no modifier `theme` in `ctx`.

// @mishanga

mishanga:
Close #861.
It's already done:
https://github.com/bem/bem-components/blob/ba59063/common.blocks/menu/menu.tmpl-specs/10-default.bemjson.js
https://github.com/bem/bem-components/blob/ba59063/common.blocks/menu/menu.tmpl-specs/10-default.html

I've just did some refactoring.

### DIFF
--- a/common.blocks/menu-item/menu-item.bemhtml
+++ b/common.blocks/menu-item/menu-item.bemhtml
@@ -2,5 +2,11 @@ block('menu-item')(
     js()(function() {
         return { val : this.ctx.val };
     }),
-    attrs()({ role : 'menuitem' })
+    attrs()({ role : 'menuitem' }),
+    match(this._menuMods).def()(function() {
+        var mods = this.mods;
+        mods.theme = mods.theme || this._menuMods.theme;
+        mods.disabled = mods.disabled || this._menuMods.disabled;
+        applyNext();
+    })
 );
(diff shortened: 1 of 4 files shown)


mishanga:
@tadatuta

common.blocks/menu-item/menu-item.bemhtml
```diff
@@ -2,5 +2,10 @@ block('menu-item')(
     js()(function() {
         return { val : this.ctx.val };
     }),
-    attrs()({ role : 'menuitem' })
+    attrs()({ role : 'menuitem' }),
+    match(this._menuMods).def()(function() {
+        this.mods.theme = this._menuMods.theme;
```
tadatuta: `var mods = this.mods;`

common.blocks/menu-item/menu-item.bemhtml
```diff
@@ -2,5 +2,11 @@ block('menu-item')(
     js()(function() {
         return { val : this.ctx.val };
     }),
-    attrs()({ role : 'menuitem' })
+    attrs()({ role : 'menuitem' }),
+    match(this._menuMods).def()(function() {
+        var mods = this.mods;
```
mishanga, line 7: @tadatuta Like this?

tadatuta, line 7: yes.

:ship:

PR #1007, opened 2014-09-09
popup: split own functionality into target modifiers (close #968)

### LINKED ISSUE #968: popup: split own functionality into target modifiers
**Update:** See comments below for discussion. Final case: https://github.com/bem/bem-components/issues/968#issuecomment-54295618.

---

First of all -- we need good name for it! ;-)

It will provide functionality about:
  1) close on Esc key
  2) correct closing within "virtual nodes tree"

dfilatov:


common.blocks/popup/popup.js
```diff
@@ -27,474 +16,64 @@ var VIEWPORT_ACCURACY_FACTOR = 0.99,
...
-
-        return this;
-    },
-
-    /**
      * Sets content
      * @param {String|jQuery} content
      * @returns {popup} this
      */
     setContent : function(content) {
```
persidskiy, line 65: Do we actually need this method in common api? Maybe move it to `_target`?

dfilatov, line 65: How are `_target` and `setContent` related to each other?

persidskiy, line 65: `setContent` is valuable because it does 2 things at the same time: updates the DOM and redraws the popup. Since we need to redraw the popup only in `_target_*` modifications, this method valuable only in this modifications. in `popup.js` it only  updates the DOM and loses value, because we can do that just through `BEMDOM.update(popup.domElem, content)`.

veged, line 65: it's inconsistent to have two ways to update content of popup (which apply differently depend on just modifier)

persidskiy, line 65: Hmm, would you implement the `setContent()` method, if you weren't need to `.redraw()` popup in the `_target_*`?

dfilatov, line 65: Yes, I would. It's often case in applications I'm developing.

persidskiy, line 65: ok

narqo:
Tests are passed! Could we squash it and merge?

dfilatov:
I don't see any reason not to merge it )

PR #1016, opened 2014-09-11
modal: make block (closes #798)

### LINKED ISSUE #798: Create modal block

persidskiy:
- [x] main idea
- [x] bh templates + tmpl-specs
- [x] specs
- [x] gemini tests


common.blocks/modal/modal.bemhtml
```diff
@@ -0,0 +1,24 @@
+block('modal')(
+
+    mix()(function() {
+        return {
+            block : 'popup',
+            js : { zIndexGroupLevel : this.ctx.zIndexGroupLevel || 19 }
```
dfilatov: 19?

common.blocks/modal/modal.js
```diff
@@ -0,0 +1,61 @@
...
+ * @class modal
+ * @bem
+ *
+ * @bemmod visible Represents visible state
+ */
+provide(BEMDOM.decl(this.name, /** @lends modal.prototype */{
+    beforeSetMod : {
+        'visible' : {
+            '' : function() {
+                var e = new events.Event('beforeClose');
```
dfilatov: Is it needed right now?

common.blocks/modal/modal.js
```diff
@@ -0,0 +1,61 @@
...
+        'js' : {
+            'inited' : function() {
+                this._popup = this.findBlockOn('popup');
+            }
+        },
+
+        'visible' : {
+            'true' : function() {
+                // Apply the animation only at first opening, otherwise the animation will be played when page loaded.
+                this.setMod('animation');
```
dfilatov: Why do you need to have modifier? Why not in css for `_visible`?

common.blocks/modal/modal.js
```diff
@@ -0,0 +1,61 @@
...
+            }
+        }
+    },
+
+    _onPopupHide : function() {
+        this.delMod('visible');
+    }
+}, /** @lends modal */{
+    live : function() {
+        this.liveInitOnBlockEvent({ modName : 'visible', modVal : '' }, 'popup', function() {
```
dfilatov: In case of you have `this._popup` it's redundant live bind.

common.blocks/modal/modal.js
narqo: :+1: for @dfilatov I don't understand the point of `_animation`

common.blocks/modal/modal.js
```diff
@@ -0,0 +1,61 @@
+/**
+ * @module modal
+ */
+
+modules.define(
+    'modal',
+    ['i-bem__dom', 'events'],
+    function(provide, BEMDOM, events) {
```
narqo: it is required to have `popup` module-dependency here. In other case, there is no guaranties that `popup` module would be resolved by the time, when `modal` would stat using it.

common.blocks/modal/_autoclosable/modal_autoclosable.js
```diff
@@ -0,0 +1,35 @@
...
+
+            '' : function() {
+                this.__base.apply(this, arguments);
+                this.unbindFrom('pointerclick', this._onPointerClick);
+            }
+        }
+    },
+
+    _onPointerClick : function(e) {
+        !dom.contains(this.elem('content'), $(e.target)) && this.delMod('visible');
```
narqo: It could be changed in "or" way: `one || two` instead of `!one && two`

common.blocks/modal/modal.tests/simple.blocks/test/test.js
```diff
@@ -0,0 +1,59 @@
...
+                });
+                this.findBlockOn('remove-text', 'button').on('click', function() {
+                    text.find('p:last').remove();
+                });
+                this.findBlockOn('wider', 'button').on('click', function() {
+                    content.css('width', parseInt(content.css('width'), 10) + 100);
+                });
+                this.findBlockOn('narrower', 'button').on('click', function() {
+                    content.css('width', parseInt(content.css('width'), 10) - 100);
+                });
```
narqo: The code above could be chained

common.blocks/modal/modal.js
```diff
@@ -0,0 +1,61 @@
...
+        this.delMod('visible');
+    }
+}, /** @lends modal */{
+    live : function() {
+        this.liveInitOnBlockEvent({ modName : 'visible', modVal : '' }, 'popup', function() {
+            this._onPopupHide();
+        });
+    }
+}));
+});
```
narqo, line 47: Line break here is missing?

common.blocks/modal/_autoclosable/modal_autoclosable.js
```diff
@@ -0,0 +1,35 @@
...
+ * @exports
+ * @class modal
+ * @bem
+ */
+provide(Modal.decl({ modName : 'autoclosable', modVal : true }, /** @lends modal.prototype */{
+    onSetMod : {
+        'visible' : {
+            'true' : function() {
+                this.__base.apply(this, arguments);
+                this.bindTo('pointerclick', this._onPointerClick);
```
veged: maybe better to `.bindToWin`? I know that modal should be resized to whole window, but... life may be vary ;-)

common.blocks/modal/modal.tests/simple.blocks/test/test.js
veged: @narqo no, there is different `button` blocks

veged: I seggest to write `function bindToButtonClick(elem, fn) { _this.findBlockOn(elem, 'button').on('click', fn) }` and use it 4 times

narqo:
JFI(@Coltspb): #968 have been merged to v2 ;)

common.blocks/modal/modal.js
persidskiy: Start from the far. We have 2 animations: for opening and for closing. The opening animation applies to the modal when it gains `_visible` modifier. But the closing animation should be applied by default. It allows us just remove `_visible` to play closing animation. 
Ok, all is well. 
But also we should set `visibility: hidden` to the modal only when the closing animation ends. Otherwise the animation won't be played. To do this, we add [visibility:visible](https://github.com/bem/bem-components/blob/27b5825cb1c284dedcd8ad2c42ddcb55c1c2a2a2/design/common.blocks/modal/_theme/modal_theme_normal.styl#L59) for animation time, which overrides `visibility: hidden` derived from regular styles. Thus, when the modal initializes, it has `visibility: hidden` and applied closing animation, which sets `visibility: visible` and plays. To prevent this, I've added `_animation` which applies animations to the modal when it opens first time. It works like `_direction_*` modifier in `popup`, which applies animations to it and don't set before first opening.

I don't like this mod, but have no idea how to do this with elegance.

common.blocks/modal/_autoclosable/modal_autoclosable.js
persidskiy: It will be closed if you click body, because mixed popup has `_autoclosable` and modal listen his closing.

persidskiy:
:up:

common.blocks/modal/modal.bemhtml
```diff
@@ -0,0 +1,24 @@
...
+    js()(true),
+
+    content()(function() {
+        return {
+            elem : 'table',
+            content : {
+                elem : 'cell',
+                content : {
+                    elem : 'content',
+                    content : this.ctx.content
```
tadatuta: `this.ctx.content -> applyNext()` ?

common.blocks/modal/_autoclosable/modal_autoclosable.bemhtml
```diff
@@ -0,0 +1 @@
+block('modal').mod('autoclosable', true).mix()({ block : 'popup', mods : { autoclosable : true } });
```
tadatuta: looks like `zIndexGroupLevel` is missing here?

common.blocks/modal/_autoclosable/modal_autoclosable.js
veged: ok than

common.blocks/modal/modal.bemhtml
veged: sure!

common.blocks/modal/modal.tests/simple.blocks/test/test.js
```diff
@@ -0,0 +1,53 @@
+modules.define('test', ['i-bem__dom'], function(provide, BEMDOM) {
+
+provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
```
sipayRT: `/** @lends popup.prototype */` confusing

common.blocks/modal/modal.js
persidskiy: @mishanga Do you have any suggestions?

persidskiy:
@mishanga check markup plz :)

persidskiy:
Guys?

narqo:
There is another problem that we haven't discuss yet: `modal` should prevent focus to go outside of the modal window. There is an article about that on [Smashing Magazine](http://www.smashingmagazine.com/2014/09/15/making-modal-windows-better-for-everyone/)

persidskiy:
@narqo We mentioned that when I was starting to collect bugs of current popup in islands. But we've decided, that cycling focus isn't the area of responsibility of popup, because it can contain not focusable elements.

common.blocks/modal/modal.js
narqo: May be we can use some non-boolean "value" for `_visibility` (e.g. `_visibility_progress`)? 

BTW, it would be much better to start use `transitionend` with `display: none` in CSS, instead of permanent `visibility: hidden`.

persidskiy: We don't have `_visibility` modifier. If you about `_visible_progress`, then how should it work? 
AFAIK we can't set new value to modifier in his `onSetMod`-handler. 

Usage of `animationend` looks good, but we will be forced to use two css-classes for animation: first for setting/unsetting `display:block` and second for setting/unsetting `animation-name`.

common.blocks/modal/modal.bemhtml
```diff
@@ -0,0 +1,24 @@
+block('modal')(
+    js()(true),
+
+    mix()(function() {
+        return {
+            block : 'popup',
+            js : { zIndexGroupLevel : this.ctx.zIndexGroupLevel || 20 },
```
narqo, line 7: Let's move this to the `Modal~getDefaultParams()`!

common.blocks/modal/modal.js
narqo: @Coltspb I still don't understand this comment:

>  Apply the animation only at first opening, otherwise the animation will be played when page loaded.

What's wrong with applying CSS animation (`animation-name` property) for `modal_js_inited` selector. See #1028 for code

common.blocks/modal/modal.bemhtml
persidskiy, line 7: And set it to popup in runtime? We don't have api for that. It's better to keep `block.params` in readonly, isn't it?

narqo, line 7: I don't understand what the "runtime" you're talking about? Blocks's `this.params` is 

``` js
this.params = objects.extend(this.getDefaultParams(), params);
```

So there shouldn't be any reasons to leave magical `|| 20` part inside the template

common.blocks/modal/modal.js
narqo: Anyway, `animation` (AFAIU, `_has-animation` is much more consistent name) is compleatly specific for `_theme_normal`. So I'm strongly insist that it mustn't be inside common block definition!

common.blocks/modal/modal.bemhtml
persidskiy, line 7: We should set `zIndexGroupLevel` to popup from modal, so I set it in template.  Otherwise, it will look that: 

``` js
// modal.js
onSetMod: {
   'js': {
        'inited': function() {
            this._popup = this.findBlockOn('popup');
            this._popup.params.zIndexGroupLevel = 20;
        }
    }
}
```

For me, it's not good practice to set params after initialization. So, I've used traditional way to set params to block - in template.

// Runtime in prev. comment means "after initialization"

narqo, line 7: Crap it's a `popup`. I don't understand why do we use different ways to set block parameters inside our own library (`js` params vs BEMJSON fields) – "this API smells" :( We need to think about it separately.

narqo:
@Coltspb how should one update content of the `modal` in the runtime? 

I have _a view block_ that is created dynamically during application runtime. The lifecycle of my application activity looks like this:

1) create html of the `view` ("view" is mixed with "modal")
2) pass `spin` inside the modal (through the modal API) and start some "service loader"
3) show the `modal`
4) after "service loader" gets data, create suitable "Model" for this data and build new content of the "view"
5) pass content from (4) to the view (replace `spin` with new content)

I want to have some king of high-level abstraction View class on top of the low-level `modal` block, so I don't want to use API of the `modal` directly from my Activitiy (_presenters_) layer.

**UPD**: I can use `bemDom.update(this._modal.elem('content'), content);` from my "view" block, but I don't think that it's very intuitive, 'cause I don't know anything about `modal__content` when looking at BEMJSON of the `modal` block ;)

persidskiy:
:up:

common.blocks/modal/modal.styl
```diff
@@ -0,0 +1,32 @@
+.modal
+{
+    position: fixed;
+    left: 0;
+    top: 0;
+    bottom: 0;
+    right: 0;
```
sipayRT:
```
    top: 0;
    right: 0;
    bottom: 0;
    left: 0;
```

common.blocks/modal/modal.styl
```diff
@@ -0,0 +1,32 @@
...
+    text-align: center;
+    overflow-y: auto;
+    -webkit-overflow-scrolling: touch;
+
+    &__table
+    {
+        display: table;
+        width: 100%;
+        height: 100%;
+        text-align: center;
```
sipayRT, line 21: в общем, прогони через CSScomb, плз

design/common.blocks/modal/_theme/modal_theme_normal.styl
```diff
@@ -0,0 +1,109 @@
...
+        background-color: #fff;
+        border: 1px solid #bfbfbf;
+    }
+
+    &.modal_js_inited
+    {
+        display: block;
+    }
+
+    &.modal_has-animation
```
mishanga: Обсудили голосом, что вместо дополнительного модификатора можно повесить анимацию на `.page > .modal...`.
Чревато тем, что кто-нибудь может положить modal сразу внутрь body на первый уровень.

common.blocks/modal/modal.styl
persidskiy, line 21: AFAIK, csscomb can't handle stylus, but I'll try.

design/common.blocks/modal/_theme/modal_theme_normal.styl
persidskiy: @narqo your opinion is very important for us =)

common.blocks/modal/modal.styl
mishanga, line 21: 1. npm i csscomb@3
2. mv *.styl *.scss
3. csscomb *.scss
4. mv *.scss *.styl
5. ???????
6. profit

design/common.blocks/modal/_theme/modal_theme_normal.styl
narqo: Looks cool! But as @mishanga said it would break in cases where `modal` is a child of `page` from it's start (for example there could be an example of block usage).

design/common.blocks/modal/_theme/modal_theme_normal.styl
```diff
@@ -0,0 +1,109 @@
+.modal_theme_normal
+{
+    display: none;
+    visibility: hidden;
+
+    background: rgba(50, 50, 50, .2);
+
+    &, & .modal__content
```
mishanga:
```
&,
& .modal__content
```

design/common.blocks/modal/_theme/modal_theme_normal.styl
```diff
@@ -0,0 +1,109 @@
...
+    &.modal_js_inited
+    {
+        display: block;
+    }
+
+    &.modal_has-animation
+    {
+        animation-name: modal_theme_normal;
+
+        .modal__content {
```
mishanga: break before brace

design/common.blocks/modal/_theme/modal_theme_normal.styl
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    100%
+    {
+        transform: scale(1.5);
+    }
+}
+
+// opening modal__content
+@keyframes modal_theme_normal_visible__content {
```
mishanga: break before brace

design/common.blocks/modal/_theme/modal_theme_simple.styl
```diff
@@ -0,0 +1,16 @@
...
+
+    &.modal_visible
+    {
+        display: block;
+    }
+
+    .modal__content
+    {
+        background: #fff;
+        box-shadow: 0 0 0 1px rgba(0,0,0,.1);
```
mishanga: spaces after commas

design/common.blocks/modal/_theme/modal_theme_normal.styl
persidskiy: So, what is better for you? Robust approach with modifier, or cool approach with some special effects.

narqo: For me, additional modifier is better: at least it doesn't break my current app ;)

persidskiy: Ok

persidskiy:
@narqo Could you, please, validate latest commits?
... and could I merge it?

aristov:
resolves https://github.com/bem/bem-components/issues/798

narqo:
@Coltspb could update the pr to satisfy latest `v2` head (e.g. #1038)?

persidskiy:
@narqo done

dfilatov:
LGTM :+1:

narqo:
@Coltspb please squash commits into one before merge

dfilatov:
@Coltspb Is there any progress?

common.blocks/modal/modal.styl
```diff
@@ -0,0 +1,37 @@
...
+    left: 0;
+
+    height: 100%;
+    width: 100%;
+
+    overflow-y: auto;
+
+    -webkit-overflow-scrolling: touch;
+
+    &__table
```
dfilatov, line 14: Could we use more modern technics (flexbox) with fallback to IE9?

persidskiy, line 14: We can try, but I haven't tested it in details. Also in this task thought about our current production, where we need to support IE8.
I'll test this deeper.

dfilatov, line 14: bem-components doesn't support ie8 at all

common.blocks/modal/modal.tmpl-specs/20-autoclosable.bemjson.js
```diff
@@ -0,0 +1,7 @@
+({
+    block : 'modal',
+    mix : { block : 'awesome' },
+    mods : { autoclosable : true, theme : 'normal' },
```
narqo: `normal` → `islands`

narqo:
@Coltspb Why there still some `_theme_normal` examples?

gemini/modal.gemini.js
```diff
@@ -0,0 +1,19 @@
...
+    root.setUrl('desktop.tests/modal/gemini/gemini.html');
+
+    gemini.suite('normal', function(suite) {
+        suite
+            .setCaptureElements('html') // because modal acts on entire window
+            .capture('plain')
+            .capture('visible', function(actions, find) {
+                var link = find('.link');
+                actions
+                    .focus(link) // because link may not get focus outline in firefox in saucelabs when we just click it
```
persidskiy, line 14: all is ok in gemini since I've added focusing manually

tavriaforever:
ui!! bem-forum was waiting for this merge! Thanks!

PR #1028, opened 2014-09-19
modal: Get rid of animation modifier

narqo:
..with additional tiny refactoring

### DIFF
--- a/common.blocks/modal/modal.js
+++ b/common.blocks/modal/modal.js
@@ -22,18 +22,8 @@ provide(BEMDOM.decl(this.name, /** @lends modal.prototype */{
             }
         },
 
-        'visible' : {
-            'true' : function() {
-                // Apply the animation only at first opening, otherwise the animation will be played when page loaded.
-                this.setMod('animation');
-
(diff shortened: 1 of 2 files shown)


common.blocks/modal/modal.js
```diff
@@ -22,18 +22,8 @@ provide(BEMDOM.decl(this.name, /** @lends modal.prototype */{
...
-
-                this._popup.setMod('visible');
-
-            },
-
-            '' : function() {
-                this._popup.delMod('visible');
-            }
+        'visible' : function(modName, modVal) {
+            this._popup.setMod(modName, modVal);
```
persidskiy, line 26: :+1:

persidskiy:
The `test` block is lazy. Try to initialize modal, but don't open it.

``` js
modules.require(['jquery'], function($){$('.modal_theme_normal').bem('modal');} );
```

narqo:
Finally, I've understood the problem with animations, thanks. `_has-animation` is the only CSS based solution that I can think of now :(

PR #1032, opened 2014-09-23
attach: synchronize _focused with inner button  #1014

hcodes:


tadatuta:
looks good but I'd rename generic `el` to more specific `mod`

hcodes:
Fixed.

aristov:
resolves https://github.com/bem/bem-components/issues/1014

hcodes:
#1059

PR #1039, opened 2014-09-29
popup: add deps for _target_position & _target_anchor

persidskiy:
Because `popup_target` declaration can fall below the `popup_target_position` or `popup_target_anchor` declarations after build. 

//cc @narqo

### DIFF
--- a/common.blocks/popup/_target/popup_target_anchor.deps.js
+++ b/common.blocks/popup/_target/popup_target_anchor.deps.js
@@ -1,0 +1,3 @@
+({
+    mustDeps : [{ mods : { target : true } }]
+})
--- a/common.blocks/popup/_target/popup_target_position.deps.js
+++ b/common.blocks/popup/_target/popup_target_position.deps.js
@@ -1,0 +1,3 @@
+({
+    mustDeps : [{ mods : { target : true } }]
+})


narqo:
Is this a problem from real life or from the theory? In general this can happen with any BEM item, not modifiers only. 

The other question is why `popup` only? At least `menu` and `select` should have the same "problem"

persidskiy:
From real life. Today I faced with it.
You can see this pr https://github.com/bem/bem-components/pull/1036 and [gemini report for it](http://bem.github.io/reports/2994/gemini-report/index.html). There the popup was broken exactly because `popup_target` declaration fell below `popup_target_anchor` declaration.

PS I know about menu and select. Fix it in this pr, or send another couple?)

veged:
maybe we should add such deps (with `include: false`) automatically in deps tech...

tadatuta:
there's no support for `include: false` in ENB :(

persidskiy:
Also, in `menu` block, the `menu_mode.js` doesn't come with `menu_mode_*.js`. 
Maybe it was broken in enb? 

//cc @andrewblond

veged:
@tadatuta I mean some "auto magic" that works like `include: false` (implemented in both deps techs: bem-tools and ENB)

regardless of this we need `include: false` in ENB

tadatuta:
@veged do we have valid cases for `include: false` if current bug will be fixed with the fact that `block_mod` will always go before `block_mod_val`?

veged:
@tadatuta yep, a lot of them — for example when we need to declare particular order between different modifiers but without enforcing dependency between them

PR #1052, opened 2014-10-15
input: make _readonly modifier (close #369)

### LINKED ISSUE #369: input: make _readonly modifier
It possible now to set HTML-attributes for controls just with full form of bemjson.

// cc @aliosv

aristov:
resolves #369

### DIFF
--- a/common.blocks/input/_readonly/input_readonly.bemhtml
+++ b/common.blocks/input/_readonly/input_readonly.bemhtml
@@ -2,4 +2,4 @@ block('input').mod('readonly', true)(
     elem('control').attrs()(function() {
         return this._.extend(applyNext(), { readonly: 'readonly' });
     })
-)
+);
--- a/common.blocks/input/_readonly/input_readonly.bemhtml
+++ b/common.blocks/input/_readonly/input_readonly.bemhtml
@@ -1,0 +1,5 @@
+block('input').mod('readonly', true)(
+    elem('control').attrs()(function() {
+        return this._.extend(applyNext(), { readonly: 'readonly' });
+    })
+)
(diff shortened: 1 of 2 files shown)


aristov:
// cc @dfilatov It works!

common.blocks/input/_readonly/input_readonly.js
```diff
@@ -0,0 +1,20 @@
+modules.define('input', function(provide, Input) {
+
+provide(Input.decl({ modName : 'readonly' }, {
+    beforeSetMod : {
+        readonly : {
+            '' : function() {
+                this.elem('control').removeAttr('readonly');
```
dfilatov, line 7: Oh, there is wrong part. You shouldn't use `beforeSetMod` for any actions. Only for preventing mod setting.

aristov, line 7: Why?

common.blocks/input/_readonly/input_readonly.bemhtml
```diff
@@ -0,0 +1,5 @@
+block('input').mod('readonly', true)(
+    elem('control').attrs()(function() {
+        return this._.extend(applyNext(), { readonly: 'readonly' });
```
narqo, line 3: `this._.extend` → `this.extend`

`this._` — is a vestige thing

common.blocks/input/_readonly/input_readonly.js
dfilatov, line 7: Because you have unpredictable behaviour in case if you have another `beforeSetMod` -> `readonly` -> `''` which returns `false`.

PR #1117, opened 2014-11-24
popup: Wrong popup position while BODY has offset (close #1111)

### LINKED ISSUE #1111: popup: Wrong popup position while BODY has offset
If `body` has some offset (e.g. `left: 20px;` in CSS), so popup's position will be wrong. 
Video(swf) - https://yadi.sk/d/lZhb6aDNcqndM

When we calculate popup left position in js, we get right number. But when we add this number to css, browser start  count not from screen position, but from `body` position. So popup's css-property `left` will be, for example, 100px, but in js method offset() will return 120px (because of `body` offset.left : 20px;)
I wrote some "костыль"(sorry for my english) for fixing this bug in `_calcTargetDimensions` method (popup_target_anchor.js)

```
var bodyOffset = this.findBlockOutside('page').domElem.offset();

if(bodyOffset.left !== 0 || bodyOffset.top !== 0) {
    anchorPos.top -= bodyOffset.top;
    anchorPos.left -= bodyOffset.left;
}
```

But I think you know better way

dfilatov:


dfilatov:
cc @narqo @veged

narqo:
looks good

sipayRT:
:+1:

PR #1139, opened 2014-11-26
popup: closes when user clicks button or some inputs via spacebar (close #1134)

### LINKED ISSUE #1134: popup: closes when user clicks button or some inputs via spacebar
The native `<input />` with some types (`button`, `reset`, `submit`, `image`, `checkbox`, `radio`, `file`) and  the `<button></button>` generate `click` event when user press spacebar on it. This click bubbles  to document and closes all popups

persidskiy:
close #1134
// cc @narqo @dfilatov

### DIFF
--- a/common.blocks/popup/_autoclosable/popup_autoclosable.js
+++ b/common.blocks/popup/_autoclosable/popup_autoclosable.js
@@ -41,8 +41,8 @@ provide(Popup.decl({ modName : 'autoclosable', modVal : true }, /** @lends popup
         if(this.hasMod('target', 'anchor') && dom.contains(this._anchor, $(e.target)))
             return;
 
-        this._inPopupPointerPress?
-           this._inPopupPointerPress = null :
+        this._preventHideByClick?
+           this._preventHideByClick = null :
            this.delMod('visible');
     }
 }, /** @lends popup */{
--- a/common.blocks/popup/popup.js
+++ b/common.blocks/popup/popup.js
@@ -45,14 +45,14 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
                 this
                     ._captureZIndex()
                     ._bindToParentPopup()
-                    .bindTo('pointerpress', this._onPointerPress);
+                    .bindTo('pointerpress pointerclick', this._setPreventHideByClick);
             },
 
             '' : function() {
                 this
                     ._releaseZIndex()
                     ._unbindFromParentPopup()
-                    .unbindFrom('pointerpress', this._onPointerPress);
+                    .unbindFrom('pointerpress pointerclick', this._setPreventHideByClick);
             }
         }
     },
@@ -76,10 +76,10 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
         return res;
     },
 
-    _onPointerPress : function() {
+    _setPreventHideByClick : function() {
         var curPopup = this;
         do {
-            curPopup._inPopupPointerPress = true;
+            curPopup._preventHideByClick = true;
         } while(curPopup = curPopup._getParentPopup());
     },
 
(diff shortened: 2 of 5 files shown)


common.blocks/popup/popup.js
```diff
@@ -45,6 +46,7 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
                 this
                     ._captureZIndex()
                     ._bindToParentPopup()
+                    .bindTo('keydown', this._onKeyDown)
                     .bindTo('pointerpress', this._onPointerPress);
```
narqo: Couldn't we add `pointerclick` listener here as well?

common.blocks/popup/popup.js
```diff
@@ -76,10 +79,23 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
         return res;
     },
 
+    _onKeyDown : function(e) {
+        var nodeName = e.target.nodeName.toLowerCase();
+
+        e.keyCode === keyCodes.SPACE && (
+            nodeName === 'button' ||
+            (nodeName === 'input' && ~INPUT_TYPES_TRIGGER_CLICK.indexOf($(e.target).attr('type')))
```
narqo: May be we should add this to `bem-core/dom`?

common.blocks/popup/popup.js
persidskiy: No, https://github.com/bem/bem-components/pull/900/files

common.blocks/popup/popup.js
persidskiy: I think it's very specific. At least, I was faced with it only here.

common.blocks/popup/popup.js
narqo: I'm not talking about removing `pointerpress`, but about adding `pointerclick` listener for cases where `click` event generated programmatically. It seems it should solve our problem in more general way. 

Of course current `_onPointerPress` should have some small changes too (and not to run two `while` loops for ppress and click).

persidskiy: Hmm, it's good idea! :)

persidskiy:
:arrow_up:

common.blocks/popup/popup.js
```diff
@@ -76,10 +76,10 @@ provide(BEMDOM.decl(this.name, /** @lends popup.prototype */{
         return res;
     },
 
-    _onPointerPress : function() {
+    _preventHideByClick : function() {
         var curPopup = this;
         do {
-            curPopup._inPopupPointerPress = true;
+            curPopup._needPreventHideByClick = true;
         } while(curPopup = curPopup._getParentPopup());
```
narqo, line 83: Maybe we could optimise `_preventHideByClick` method in some way? Now this loop runs twice on each click: first while handling `pointerpress`, second on `pointerclick`.

persidskiy, line 83: Yep, but it shouldn't be expensive. And it more robust.

common.blocks/popup/_autoclosable/popup_autoclosable.js
```diff
@@ -41,8 +41,8 @@ provide(Popup.decl({ modName : 'autoclosable', modVal : true }, /** @lends popup
         if(this.hasMod('target', 'anchor') && dom.contains(this._anchor, $(e.target)))
             return;
 
-        this._inPopupPointerPress?
-           this._inPopupPointerPress = null :
+        this._needPreventHideByClick?
```
dfilatov: Use either `_preventHideByClick` or `_needToPreventHideByClick`. I like first one.

persidskiy:
@dfilatov 

```
_preventHideByClick() -> _setPreventHideByClick()
_needPreventHideByClick -> _preventHideByClick
```

dfilatov:
@Coltspb ok

persidskiy:
Then, what about merge?

dfilatov:
lgtm

persidskiy:
@narqo ?

narqo:
:+1:

PR #1149, opened 2014-11-26
Used BH only to build sets

### LINKED ISSUE #1148: sets: build BH only

blond:
Resolved #1148 
1. Для сборки html на сервере теперь используется `?.bh.js` вместо `?.bemhtml.js`.
2. Для клиентского JavaScript используется `?.browser.bh.js` вместо `?.browser.bemhtml.js`.

### DIFF
--- a/.enb/make.js
+++ b/.enb/make.js
@@ -14,11 +14,10 @@ var DEFAULT_LANGS = ['ru', 'en'],
     js = require('enb-diverse-js/techs/browser-js'),
     ym = require('enb-modules/techs/prepend-modules'),
     bemhtml = require('enb-bemxjst/techs/bemhtml-old'),
-    html = require('enb-bemxjst/techs/html-from-bemjson'),
+    html = require('enb-bh/techs/html-from-bemjson'),
     bh = require('enb-bh/techs/bh-server'),
     bhServerInclude = require('enb-bh/techs/bh-server-include'),
     bhYm = require('enb-bh/techs/bh-client-module'),
-    bhHtml = require('enb-bh/techs/html-from-bemjson'),
     copyFile = require('enb/techs/file-copy'),
     mergeFiles = require('enb/techs/file-merge'),
     mergeBemdecl = require('enb-bem-techs/techs/merge-bemdecl'),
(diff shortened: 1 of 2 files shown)


tadatuta:
lgtm. gemini fails for some reason on `select`'s tests but it's not connected with PR

blond:
I think the problem is the lack of `select` implementation in bh tech for gemini tests: https://github.com/bem/bem-components/blob/v2/common.blocks/select/select.tests/gemini.blocks/select/select.bemhtml.

narqo:
In general we've started development of the new block from creating a test bemjson for it and writing bemhtml templates. AFAII this habit would be broken.

1) Why?
2) We (at least me and @dfilatov) still use bem-server to build test bundles? Would the behaviour of `enb make tests desktop.tests` and `bem make desktop.tests` still be the same?

tadatuta:
1. Main reason is to reduce test time in travis
2. Nope. But that can save your habit with BEMHTML first approach for now ;)

dfilatov:
It's too strange If reducing test time in travis more important than the ease of development.

veged:
Which time reduction we talking about (in ms and in %)?

I think we can't do this before `enb` will be able to make dev-server.

blond:
> Which time reduction we talking about (in ms and in %)?

`npm run build` — 323512ms vs 75173ms ≈ 4X faster

> I think we can't do this before enb will be able to make dev-server.

WIP — https://github.com/bem/bem-components/pull/1153

blond:
@dfilatov, @narqo now you can use environment variable `BEM_TEMPLATE_ENGINE` with value `BEMHTML`  to build sets by `BEMHTML` instead of `BH`.

blond:
@tadatuta need review

qfox:
@andrewblond Don't you think that better use BEMHTML by default as it was and use BH template engine (by setting env variable) at travis?

.enb/make.js
```diff
@@ -191,39 +192,45 @@ module.exports = function(config) {
...
             // Build htmls
-            nodeConfig.addTechs([
-                [html],
-                [bhHtml, { target : '?.bh.html' }]
+            nodeConfig.addTechs(BEM_TEMPLATE_ENGINE === 'BH'? [
+                [bh, {
+                    jsAttrName : 'data-bem',
+                    jsAttrScheme : 'json'
+                }],
+                [bhHtml]
```
tadatuta, line 230: if we agreed that `BH` should mimic to `BEMHTML` can't we use the same tech for both engines to create HTML?
the rest looks good for me so if it's hard to implement with `ENB` — squash commits!

blond, line 230: > if we agreed that BH should mimic to BEMHTML can't we use the same tech for both engines to create HTML?

Now BH can mimic to BEMHTML for client techs only.

PR #1163, opened 2014-12-02
Add select.setOptions method to dynamic update items #1100

### LINKED ISSUE #1100: select: allow dynamic update items

slonoed:
This resolves #1100

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -11,7 +11,8 @@
         "*.pages/**",
         "**/*.i18n/**",
         "libs/**",
-        "node_modules/**"
+        "node_modules/**",
+        "dist"
     ],
     "fileExtensions": [".js", ".bemtree", ".bemhtml"],
     "requireSpaceAfterKeywords": ["do", "else"],
--- a/common.blocks/menu/menu.spec.js
+++ b/common.blocks/menu/menu.spec.js
@@ -196,6 +196,13 @@ describe('menu', function() {
                 .domElem.html().should.be.equal('content');
         });
     });
+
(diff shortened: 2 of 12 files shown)


aristov:
I run command `enb make specs desktop.specs/select` and get: `Error: Module "select": can't resolve dependence "menu"`
Addition `menu` block to `deps`-file resolves this problem.

slonoed:
@aristov 
`npm run gemini`
works well localy, but fail in travis. Can you check and merge PR?

aristov:
cc @veged, @dfilatov

qfox:
Do we need `getOptions` method? It uses `menu.ptp.getItems`. Are you sure about naming?

slonoed:
@zxqfox, `.getOptions` - bad idea. Only info than select should return - is selected values.

qfox:
@SLonoed yeah, but that's strange to have `setOptions` without getter. Do you thought about appending some additional options to select? Anyway, it was just a suggestion.

slonoed:
@zxqfox, maybe it about naming. `.updateOptions` can be more consistent.

narqo:
Hey, @SLonoed thanks for pr and sorry for delay.

After a (long) discussion \w @veged and @dfilatov we decided that re-rendering the whole select is much better (and much universal) solution for such cases. At least on the current API state. Currently `updateOptions` doesn't have any particular benefits. But it brings the requirement of embedding BEMHTML templates inside client JS, as well as additional nested element.

Exact problem could be solve with such API enhancements on the project's level.

PR #1165, opened 2014-12-03
checkbox: invisible check click

incrop:
You can set checkbox on by clicking slightly above it.
That is because `.checkbox__box::after` is transparent, but still visible for clicks.
This fix sets `visibility` attribute in addition to `opacity`.


hcodes:
\cc @mishanga

mishanga:
ok

PR #1177, opened 2014-12-06
modal: Styles of the block could be overridden with popup styles

### LINKED ISSUE #1176: modal: Styles of the block could be overridden with popup styles
Currently `modal` is shoulddepsing `popup` :) So there is no guaranty that `popup`'s styles won't override `modal`'s

2014-12-06 17-07-34 vertis_pilot_admin

narqo:
fix #1176


persidskiy:
:+1:

PR #1183, opened 2014-12-08
menu: no bottom padding with scroll

### LINKED ISSUE #1158: menu: bottom padding is absent in case of specified height and scroll
Height is not specified:
normal

Height is specified:
broken

mishanga:
fix #1158


dfilatov:
:+1:

PR #1191, opened 2014-12-14
[WIP] tabindex: make sure that all controls can have it

qfox:


tadatuta:
// cc @innabelaya

qfox:
@tadatuta need to update docs?

tadatuta:
yes, but current version of documentation is not merged and still in a [branch](https://github.com/bem/bem-components/pull/1123)

qfox:
@tadatuta np. I just looked at artistov/v2-form and I trying to prepare core sources to be consistent to make easier work on form functionality.

Should I add these fields to first example?

innabelaya:
`id` and `tabIndex` fields are added to the block description (#1123)

narqo:
> I just looked at artistov/v2-form and I trying to prepare core sources to be consistent to make easier work on form functionality

Are there any real cases where anyone needs ids in checkboxes? 

@tadatuta Shouldn't we test somehow such enhancements with our testers? For example, how `checkbox_type_button` would work here? How `label[id = 'id1'] + checkbox_type_button[id = 'id1']` would  behave?

qfox:
@narqo I've added this because I can't pass id to it. But for now I don't feel we really need it and I dropped passing id.

I propose we shouldn't block `tabIndex` functionality because of controversial `id`-passing functionality.

qfox:
@tadatuta If you think we need this `id`-passing I'll make another PR with it.

narqo:
@zxqfox what about `checkbox_type_button`? Does `tabIndex` works there? :)

qfox:
@narqo ;-(

narqo:
Another thing I've noticed with @dfilatov's help: how this supposed to work from `checkbox-group`?

qfox:
@narqo Now how ;-)

Already looking to it.

qfox:
@narqo Should we solve this problem here? Can we left `-group` as is for now?

qfox:
I know what I want. Templates mixing for passing tabIndex ;-)

qfox:
@narqo https://github.com/zxqfox/bem-components/commit/38bf6cf192128385e59ac142ab91f2a66073be5c Ok?

narqo:
I think the code is fine now, thanks.

@tadatuta as you opened #1206 I suggest to see on this issue (as on the whole `tabIndex` API in the library) from the Accessibility point of view. I don't have any expertise here, the only think that I know is that tabindexes other than `0` or `-1` shouldn't be used in the static markup.

qfox:
Okay. Time to cleanup. Thanks!

common.blocks/radio-group/radio-group.bemhtml
```diff
@@ -27,6 +27,7 @@ block('radio-group')(
                     val : option.val,
                     text : option.text,
                     title : option.title,
+                    tabIndex : option.tabIndex || ctx.tabIndex,
```
veged, line 30: there is should not be separate tabIndex on options — just common tabIndex from parent

veged:
it's better to reformulate title of PR and issue because difference not only about checkbox

common.blocks/radio-group/radio-group.bemhtml
qfox, line 30: You just didn't took a look at my test bundle. E.g. `flex: reverse` is the case.

qfox:
@veged Agreed. Feel free to do it!

veged:
@zxqfox you know better what exactly was affected

common.blocks/radio-group/radio-group.bemhtml
veged, line 30: OMG! looks like very odd case, but since it doesn't take a lot of code complexity maybe we can leave it ;-)

qfox:
@veged checkbox/radio/checkbox-group/radio-group.

Rebased.

common.blocks/radio-group/radio-group.bemhtml
qfox, line 30: Well, yes ;-) It would be nice.

common.blocks/checkbox-group/checkbox-group.bemhtml
```diff
@@ -29,6 +29,7 @@ block('checkbox-group')(
                     val : option.val,
                     text : option.text,
                     title : option.title,
+                    tabIndex : option.tabIndex || ctx.tabIndex,
```
narqo, line 32: Looks like there is no need to have separate tabindex in each option. So `ctx.tabIndex` should be enough.

common.blocks/dropdown/dropdown.ru.md
```diff
@@ -289,3 +290,34 @@
...
+}
+```
+
+```js
+{
+    block : 'dropdown',
+    mods : { switcher : 'button', theme : 'islands', size : 'm' },
+    switcher : {
+        block : 'button',
+        tabIndex : 5,
```
narqo, line 317: It seems that this is too hardcore usecase to be shown it documentation

common.blocks/radio-group/radio-group.bemhtml
narqo, line 30: It seems that such things should be solved in other ways. If you start using tabindex'es you will have to use them (properly) within the whole page.

aristov:
@zxqfox ping

common.blocks/checkbox-group/checkbox-group.bemhtml
qfox, line 32: I'm not sure... There are cases when I need to fix tabIndex order inside a group. Can I redefine that in project level? That's not clear for me how to do that (because I'm not so familiar with bemhtml). If so I'm totally agree to left here `ctx.tabIndex` only.

common.blocks/dropdown/dropdown.ru.md
qfox, line 317: Just a more complex example. Should we drop that?

common.blocks/radio-group/radio-group.bemhtml
qfox, line 30: @narqo Yeah, probably. `tabIndexManager` or something? Is that a blocker for PR?

desktop.pages/tabindex/tabindex.bemjson.js
```diff
@@ -0,0 +1,272 @@
+({
```
qfox, line 1: Should we left here that example?

qfox:
@aristov Not sure what to do here. ;-)

aristov:
Close in favour of #1675.

Issue #125, opened 2013-07-25
Configure precommit hook for tests


veged:
If we try to commit some BEM-item with tests, we should run it in precommit hook.

mishanga:
Is it still actual?

tadatuta:
I think it is.

qfox:
@veged Should it run tests in dependants? Should it run block tests if element/modifier changed?

It will be much harder if it is and will require some special `js` tools.

qfox:
ping

deeonis:
closed in favour Travis CI

PR #1196, opened 2014-12-15
[WIP] githooks: call magic make for changed blocks at precommit

qfox:
Relates to #125


qfox:
@veged @arikon Guys, what you think about it?

narqo:
@zxqfox do we still need it since we've decided to "won't fix" #125?

qfox:
@narqo Seems like. I don't know why. Really.

PR #1199, opened 2014-12-16
progressbar: Corrections after review

gela-d:


common.blocks/progressbar/progressbar.js
```diff
@@ -23,7 +23,8 @@ provide(BEMDOM.decl(this.name, /** @lends progressbar.prototype */{
      * @param {Number} value
      */
     setVal : function(value) {
-        this._setBarWidth(this._val = value);
+        this.elem('bar').css('width', (this._val = value) + '%');
```
persidskiy, line 27: why wouldn't move `this._val = value` to separate line?

persidskiy:
Why do you check [private fields](https://github.com/bem/bem-components/blob/v2/common.blocks/progressbar/progressbar.spec.js#L24) in specs?

persidskiy:
`setVal` doesn't return `this` in `_theme_simple` (https://github.com/bem/bem-components/blob/v2/design/common.blocks/progressbar/_theme/progressbar_theme_simple.js#L20)
also there is no need to write documentation for `@override`d metrods

common.blocks/progressbar/progressbar.js
gela-d, line 27: @Coltspb why we should? :)
In this case we have minus one line.

gela-d:
> Why do you check private fields in specs?

@Coltspb Thanks, fixed.

> also there is no need to write documentation for @overrided metrods

Our method add some functionality in `progressbar_theme_simple.js`, thats why I want to write documentation for `@override` method

common.blocks/progressbar/progressbar.spec.js
```diff
@@ -21,21 +21,18 @@ describe('progressbar', function() {
             var value = 15;
             progressbar.setVal(value);
             progressbar.elem('bar')[0].style.width.should.be.equal(value + '%');
-            progressbar._val.should.be.equal(value);
         });
 
         it('should set correct percent if value is string with numbers', function() {
```
narqo: This test contravenes block's API from jsdoc: `Progressbar#setVal(val {Number})`. Why we care about string behaviour?

design/common.blocks/progressbar/_theme/progressbar_theme_simple.js
```diff
@@ -18,6 +18,7 @@ provide(Progressbar.decl({ modName : 'theme', modVal : 'simple' }, /** @lends pr
     setVal : function() {
         this.__base.apply(this, arguments);
```
narqo: You could write it in the way below, to be more accurate:

``` js
var res = this.__base.apply(this, arguments);
this.elem('text').text(this._val);
return res;
```

gela-d:
@narqo done

narqo:
looks good

PR #1201, opened 2014-12-19
Remove symlinks

### LINKED ISSUE #1200: Symlinks on Windows
На винде падает сборка стайлуса из-за симлинок.

По логу ошибки там очевидно,  что stylus рассматривает симлинку как обычный файл и не подменяет контент по пути, а просто читает его (строка, которая представляет собой путь) и конечно же не может скомпилировать.

eGavr:
closes #1200

### DIFF
--- a/common.blocks/popup/popup.tests/gemini.blocks/test-nested/test.styl
+++ b/common.blocks/popup/popup.tests/gemini.blocks/test-nested/test.styl
@@ -1,1 +1,38 @@
-../../simple.blocks/test/test.styl
+.test
+{
+    position: absolute;
+}
+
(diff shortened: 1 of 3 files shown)


tadatuta:
update with css include instead

tadatuta:
after investigation in https://github.com/bem/bem-components/pull/1410 decided to get back to copypaste solution as otherwise we must hardcode relative urls to **built** files.

narqo:
Should it be closed in favour of #1410? It seems that these two prs are about the same things.

aristov:
Close in favour of #1410

PR #1215, opened 2014-12-25
popup: doesn't invalidate zIndex when anchor set while popup is hidden (close #1214)

### LINKED ISSUE #1214: popup: doesn't invalidate zIndex when anchor set while popup is hidden
it was mine mistake when i worked on https://github.com/bem/bem-components/pull/972

persidskiy:
close #1214

### DIFF
--- a/common.blocks/popup/_target/popup_target_anchor.spec.js
+++ b/common.blocks/popup/_target/popup_target_anchor.spec.js
@@ -199,8 +199,7 @@ describe('popup', function() {
             popup2.setMod('visible');
             popup.setMod('visible');
 
-            Number(popup.domElem.css('z-index'))
-                .should.be.gt(Number(popup2.domElem.css('z-index')));
+            getZIndex(popup).should.be.gt(getZIndex(popup2));
         });
 
         it('should properly use zIndexGroupLevel param', function() {
@@ -216,8 +215,7 @@ describe('popup', function() {
             popup2.setMod('visible');
             popup.setMod('visible');
 
-            Number(popup.domElem.css('z-index'))
-                .should.be.lt(Number(popup2.domElem.css('z-index')));
+            getZIndex(popup).should.be.lt(getZIndex(popup2));
         });
 
         it('should properly use z-index-group blocks which may be in anchor parents', function() {
@@ -250,11 +248,9 @@ describe('popup', function() {
             popup2.setMod('visible');
             popup.setMod('visible');
 
-            Number(popup.domElem.css('z-index'))
-                .should.be.lt(Number(popup2.domElem.css('z-index')));
+            getZIndex(popup).should.be.lt(getZIndex(popup2));
 
-            Number(popup2.domElem.css('z-index'))
-                .should.be.lt(Number(popup3.domElem.css('z-index')));
+            getZIndex(popup2).should.be.lt(getZIndex(popup3));
         });
 
         it('should consider zIndexGroupLevel of parent popup', function() {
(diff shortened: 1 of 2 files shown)


persidskiy:
@narqo look at this pls

common.blocks/popup/_target/popup_target_anchor.spec.js
```diff
@@ -268,15 +264,18 @@ describe('popup', function() {
                     });
 
             popup.setMod('visible');
-            popup2.setAnchor(popupAnchorDomElem).setMod('visible');
 
-            Number(popup.domElem.css('z-index'))
-                .should.be.lt(Number(popup2.domElem.css('z-index')));
+            popup2.setAnchor(popupAnchorDomElem).setMod('visible');
+            getZIndex(popup).should.be.lt(getZIndex(popup2), 'initial');
```
narqo: Can you write more informative message in case the assertion fail?

narqo: I'm not sure, but I thought that  assertion `message` should contain addition information about particular error, e.g.

``` js
  › node -p "require('assert').fail(1, 0, 'one does not equal to zero')"

assert.js:92
  throw new assert.AssertionError({
        ^
AssertionError: one does not equal to zero
    at [eval]:1:19
```

From this point of view your messages in assertions look strange, sorry :)

common.blocks/popup/_target/popup_target_anchor.spec.js
```diff
@@ -300,4 +299,8 @@ function buildPopup(parentDomElem, bemjson) {
         .bem('popup');
 }
 
+function getZIndex(popup) {
```
narqo, line 322: May be `getCssZIndex`? IMHO, this name looks more clear.

common.blocks/popup/_target/popup_target_anchor.spec.js
```diff
@@ -268,15 +264,18 @@ describe('popup', function() {
...
             popup.setAnchor(popup2.domElem.find('span:first'));
+            getZIndex(popup).should.be.gt(getZIndex(popup2), 'popup moved to popup2 while visible');
 
-            Number(popup.domElem.css('z-index'))
-                .should.be.gt(Number(popup2.domElem.css('z-index')));
+            popup
+                .delMod('visible')
+                .setAnchor(popupAnchorDomElem)
+                .setMod('visible');
+            getZIndex(popup).should.be.lt(getZIndex(popup2), 'popup moved from popup2 while hidden');
```
narqo: May be we should move it to a separate test-suite? I think "popup should recalculate zIndexGroupLevel if anchor is changed" looks like a self sufficient statement in BDD terms

common.blocks/popup/_target/popup_target_anchor.spec.js
persidskiy, line 322: no difference)
Is there any z-indexes not in css?

narqo, line 322: ZIndexGroupLevel / z-index-group_level_N

persidskiy, line 322: it is z-index-group, not z-index

persidskiy:
:up:

common.blocks/popup/_target/popup_target_anchor.spec.js
narqo, line 322: :(

PR #1219, opened 2014-12-26
progressbar: rename `value` field to `val` in bemjson

### LINKED ISSUE #1218: progressbar: rename `value` field to `val` in bemjson
We should have a consistent naming over the library. According to other blocks we should rename `value` field to `val`.

gela-d:
Close #1218


dfilatov:
:+1:

sipayRT:
:+1:

PR #1221, opened 2014-12-26
Move jscs rules to preset @v2

### LINKED ISSUE #1220: update jscs and move code style to plugin
We starting work at bem-forms and need the same jscs rules. So there are at least 3 repos with the same jscs rules.

Can we move this out to separate repo with jscs plugin?

Btw. Some rules can be developed especially for .browser.js, e.g. to check single quotes inside onSetMod:

```
onSetMod: {
  'js': {
    'inited': function () { // valid
    }
  },
  another: function () { } // invalid
}
```

What you think?

/cc @veged @dfilatov

qfox:
Fixes #1220


tadatuta:
@zxqfox could you please rebase? we finally agreed to merge :)

qfox:
Done. Checking bem-core.

qfox:
bem-core: 27 code style errors found. And there are real errors that can be fixed easily.

So I'm not against merging this. ;-) Thanks!

tadatuta:
@zxqfox thank you very much!

PR #1260, opened 2015-01-16
progressbar: en version added

innabelaya:


innabelaya:
@mursya Thanks! Corrected.

Issue #1188, opened 2014-12-10
popup: Wrong popup position if BODY has css margin


hcodes:

qfox:
Related to #1006

hcodes:
Fixes #1267

PR #1267, opened 2015-01-21
 popup: Wrong popup position if BODY has css margin #1188

hcodes:


hcodes:
/cc @dfilatov 
Added gemini tests.
Tested on browsers: FF, Chrome, Opera 12.16, IE9-11.

dfilatov:
lgtm

PR #1278, opened 2015-01-23
fix copy/paste issue in checkbox_type_button.spec.js

### LINKED ISSUE #1277: copy/paste issue in checkbox_type_button.spec.js
https://github.com/bem/bem-components/blob/v2/common.blocks/checkbox/_type/checkbox_type_button.spec.js#L50
replace  disabled with focused

Guria:
resolve #1277


tadatuta:
@Guria Thanks!

PR #1284, opened 2015-01-26
Update to bem-core@2.5.1

### LINKED ISSUE #1283: Update bem-core version

tadatuta:
Closes #1283

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -11,7 +11,7 @@
     "libs"
   ],
   "dependencies": {
-    "bem-core": "2.5.0"
+    "bem-core": "2.5.1"
   },
   "devDependencies": {
     "bem-pr": "~0.10.0"


bower.json
```diff
@@ -11,7 +11,7 @@
     "libs"
   ],
   "dependencies": {
-    "bem-core": "2.5.0"
+    "bem-core": "^2.5.1"
```
dfilatov: We decided to use only strict versions.

tadatuta: @dfilatov 
There was idea of meta-library which should provide one "official" tested version scope but that's not very cool to have strict versions for base libs as it may result in impossibility for `bower` to resolve deps when there are a few top level libs using different strict versions which are actually compatible. In Islands `^` is used.

Are you sure about this decision?

dfilatov: Yes, we're, at least for a while.

PR #1306, opened 2015-01-28
Add popup examples

tadatuta:


common.blocks/popup/popup.examples/10-popup.blocks/test/test.js
```diff
@@ -0,0 +1,19 @@
...
+
+provide(BEMDOM.decl(this.name, {
+    onSetMod : {
+        'js' : {
+            'inited' : function() {
+                var link = this.findBlockInside('link'),
+                    popup = this.findBlockInside('popup')
+                        .setAnchor(link);
+
+                link.on('click', function() {
```
sipayRT, line 11: @tadatuta why not `pointerclick`?

tadatuta, line 11: @sipayRT that's normalised BEM event

Issue #1308, opened 2015-01-29
Align badges in docs


hcodes:

blond:
Use [shields.io](http://shields.io/) for badges.

hcodes:
Fixes #1309

PR #1309, opened 2015-01-29
 Align badges in docs #1308

hcodes:


blond:
Wy not `flat` style?

Need add `?style=flat` to img url. For, example: https://img.shields.io/travis/bem/bem-components/v2.svg?style=flat

hcodes:
Added.

blond:
:+1:

PR #1343, opened 2015-02-08
[wip] replaced double quoted data-bems with singled ones

qfox:
For https://github.com/bem/bem-core/issues/250
https://github.com/bem/bem-core/pull/742


qfox:
But probably we don't need this because html-differ should parse it safely.

narqo:
@zxqfox as we don't have released `bem-core@3.0` it seems that this pr is premature. Sorry

qfox:
I don't know what is it ;-)

PR #1344, opened 2015-02-09
button: Improve a11y

hcodes:


common.blocks/button/button.js
```diff
@@ -116,7 +116,7 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends butto
     },
 
     _onKeyDown : function(e) {
-        if(this.hasMod('disabled')) return;
+        if(this.hasMod('disabled') || e.shiftKey || e.ctrlKey || e.altKey || e.metaKey) return;
```
dfilatov: How are these changes related with a11y?

hcodes:
https://lego.yandex-team.ru/a11y/keyboard-management/
Модификаторы

hcodes:
Fixed.

common.blocks/button/_type/button_type_link.js
```diff
@@ -23,12 +23,16 @@ provide(Button.decl({ modName : 'type', modVal : 'link' }, /** @lends button.pro
         'disabled' : {
             'true' : function() {
                 this.__base.apply(this, arguments);
+
                 this.domElem.removeAttr('href');
+                this.domElem.attr('aria-disabled', true);
```
narqo, line 28: ~~Can't this two calls be chained?~~

narqo, line 28: 1) Why do we do this only for `button_type_link`? What about `button` itself?
2) Can we do it in the `control` base class? Does this behaviour differ from block to block?

narqo:
What about initial disabled state? What if `_disabled` was set in BEMJSON?

PR #1347, opened 2015-02-10
Replace this.isArray -> Array.isArray in bemhtml #1338

hcodes:


qfox:
I think we should we drop this method from `i-bem.bemhtml` if we don't use it.

hcodes:
bem/bem-core#872

PR #1348, opened 2015-02-10
changed the conditions of ie-styles connection

sipayRT:


sipayRT:
@andrewblond can you help with dist-config for IE?

PR #1355, opened 2015-02-13
progressbar: Improve a11y

hcodes:


common.blocks/progressbar/progressbar.bh.js
```diff
@@ -1,10 +1,18 @@
 module.exports = function(bh) {
 
     bh.match('progressbar', function(ctx, json) {
-        var val = json.val;
+        var val = json.val || 0;
 
         ctx
-            .js({ val : val })
+            .js({ val : json.val })
```
sipayRT, line 7: `json.val` --> `val`

hcodes, line 7: 0 need for the attr and inline style.

sipayRT, line 7: in this case you will get an error when using getVal(), if BEMJSON hasn't `val`.

narqo, line 7: :+1: for @sipayRT 

It's strange to have default 0 value in templates, while allow JS to fail.

common.blocks/progressbar/progressbar.bemhtml
```diff
@@ -4,11 +4,22 @@ block('progressbar')(
...
     }),
 
+    attrs()(function() {
+        return {
+            role : 'progressbar',
+            'aria-label' : this.ctx.ariaLabel,
+            'aria-labelledby' : this.ctx.ariaLabelledBy,
+            'aria-valuemin' : 0,
+            'aria-valuemax' : 100,
+            'aria-valuenow' : this.ctx.val || 0
```
narqo, line 14: According to our docs `this.ctx.val` should always [be `Number`](https://en.bem.info/libs/bem-components/v2.0.0/desktop/progressbar/docs/#val). In other cases 

1) you should fix js-mode above or block's JS will fail as @sipayRT have mentioned
2) I insist on adding tests (specs and tmpl-specs) for such cases.

PR #1356, opened 2015-02-13
input, textarea: Improve a11y

hcodes:


common.blocks/input/__control/input__control.bemhtml
```diff
@@ -8,8 +8,11 @@ block('input').elem('control')(
                 name : input.name,
                 value : input.val,
                 maxlength : input.maxLength,
+                accesskey : input.accessKey,
```
narqo: Any use cases for that?

hcodes: Search input.

PR #1357, opened 2015-02-13
dropdown, popup, modal: Improve a11y

hcodes:


common.blocks/popup/popup.tmpl-specs/10-id.bemjson.js
```diff
@@ -0,0 +1,4 @@
+({
+    block : 'popup',
+    id : 1
```
narqo: Please note: `id`s are `String`, according to [our documentation](https://en.bem.info/libs/bem-components/v2.0.0/desktop/button/docs/#buttonid) (_2^53 is too small for current web apps_)

PR #1363, opened 2015-02-17
link: Support a11y

hcodes:


common.blocks/link/link.bh.js
```diff
@@ -11,7 +11,10 @@ module.exports = function(bh) {
             attrs = {},
             tabIndex;
 
-        if(!ctx.mod('disabled')) {
+        if(ctx.mod('disabled')) {
+            ctx.js(url? { url : url } : true);
```
tadatuta, line 15: won't this line overwrite js params from bemjson?

PR #1370, opened 2015-02-19
menu, menu-item: Improve a11y

hcodes:


common.blocks/menu-item/menu-item.js
```diff
@@ -31,6 +31,17 @@ provide(BEMDOM.decl(this.name, /** @lends menu-item.prototype */{
...
                 this.__base.apply(this, arguments);
                 this.delMod('hovered');
+                this.domElem.attr('aria-disabled', 'true');
+            },
+            '' : function() {
+                this.__base.apply(this, arguments);
+                this.domElem.removeAttr('aria-disabled');
+            }
+        },
+        'checked' : {
```
narqo, line 42: code-style: no empty line between methods

common.blocks/menu-item/menu-item.spec.js
```diff
@@ -1,7 +1,9 @@
 modules.define(
     'spec',
-    ['menu-item', 'i-bem__dom', 'jquery', 'sinon', 'BEMHTML'],
-    function(provide, MenuItem, BEMDOM, $, sinon, BEMHTML) {
+    ['menu-item', 'i-bem__dom', 'jquery', 'sinon', '', 'BEMHTML'],
```
narqo: what does this `''` means?

common.blocks/menu-item/menu-item.spec.js
```diff
@@ -37,6 +39,26 @@ describe('menu-item', function() {
         });
     });
 
+    describe('disabled', function() {
+        it('should set "aria-disabled" attribute if disabled', function() {
+            menuItem.setMod('disabled', true);
+            expect(menuItem.domElem.attr('aria-disabled')).to.equal('true');
```
narqo: Please be consistence! At least within one file (e.g. look how the code have been written several lines below):
1. no need this `expect()` at this particular case;
2. we use "should"-style of BDD, so should be `menuItem.domElem.attr('aria-disabled').should.be.equal('true')`

narqo: BTW, it's a good practice to check that this particular attribute haven't been set, before first `setMod` call.

common.blocks/menu-item/menu-item.tmpl-specs/20-a11y.bemjson.js
```diff
@@ -0,0 +1,8 @@
+({
+    block : 'menu-item',
+    mods : { theme : 'islands', disabled : true, checked : true },
+    val : 1,
+    ariaLabel : 'label',
+    ariaLabelledBy : 'id0',
```
narqo: WAT? It's a `menu-item`

common.blocks/menu/menu.bemhtml
```diff
@@ -13,10 +13,12 @@ block('menu')(
                             ctx.val.indexOf(val) > -1 :
                             ctx.val === val);
                 },
-                iterateItems = function(content) {
+                iterateItems = (function(content) {
                     var i = 0, itemOrGroup;
                     while(itemOrGroup = content[i++]) {
                         if(itemOrGroup.block === 'menu-item') {
+                            itemOrGroup.attrs = (itemOrGroup.attrs || {});
+                            itemOrGroup.attrs.id = itemOrGroup.attrs.id || this.generateId(); // Support a11y
```
narqo: Can't you add this generated "id" attribute within `menu-item` template? This will give you opportunity to not bind the function to `this` here

narqo: In any case, it's so weird to pass "attrs" field here – we have custom `id` field in other blocks (see [`button` docs](https://en.bem.info/libs/bem-components/v2.0.0/desktop/button/docs/#buttonid), as well as `menu` itself). Why should `menu-item` differ?

common.blocks/menu/menu.bemhtml
```diff
@@ -44,8 +46,16 @@ block('menu')(
         });
     }),
     attrs()(function() {
-        var attrs = { role : 'menu' };
+        var ctx = this.ctx,
+            attrs = {
+                role : 'menu',
+                id : ctx.id,
+                'aria-label' : ctx.ariaLabel,
+                'aria-labelledby' : ctx.ariaLabelledBy
```
narqo, line 52: Have you found any particular use cases to these `aria-label*` attributes? Can you show or make such?

common.blocks/menu/menu.spec.js
```diff
@@ -176,6 +176,11 @@ describe('menu', function() {
             menuItems2[0].hasMod('hovered').should.be.true;
         });
 
+        it('should change "aria-activedescendant" attribute if hovered item', function() {
+            var id = menuItems[1].domElem.attr('id');
+            menuItems[1].setMod('hovered');
+            expect(menu.domElem.attr('aria-activedescendant')).to.equal(id);
```
narqo: `.should.be.equal()`

common.blocks/menu/menu.bemhtml
```diff
@@ -44,8 +46,16 @@ block('menu')(
...
+            };
+
         this.mods.disabled || (attrs.tabindex = 0);
```
narqo: Shouldn't `menu` set `aria-disabled` in this case?

common.blocks/menu/menu.js
```diff
@@ -134,8 +134,10 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
     _onItemHover : function(item) {
         if(item.hasMod('hovered')) {
             this._hoveredItem && this._hoveredItem.delMod('hovered');
+            this.domElem.attr('aria-activedescendant', item.domElem.attr('id'));
```
narqo: What if we add `MenuItem#getId()` method, that will return cached `id` property. 

``` js
// menu-item.js
onSetMod : {
  'js' : {
    'inited' : function() {
      this._id = this.domElem.attr('id');
    }
  }
},

getId : function() {
  return this._id;
}
```

Does this DOM attribute getters efficient, in case we check them in every `pointerover` handler? 

//cc @dfilatov @veged

common.blocks/menu/menu.bemhtml
narqo, line 52: BTW: [MSDN describes](https://msdn.microsoft.com/en-us/library/ie/cc848875%28v=vs.85%29.aspx) it as `ariaLabelledby` not `ariaLabelledBy`

common.blocks/menu/menu.tmpl-specs/40-a11y.html
```diff
@@ -0,0 +1 @@
+<div class="menu menu_theme_islands menu_disabled menu__control i-bem" data-bem="{&quot;menu&quot;:{}}" role="menu" aria-label="label" aria-labelledby="id0"><div class="menu-item menu-item_theme_islands menu-item_disabled i-bem" data-bem="{&quot;menu-item&quot;:{&quot;val&quot;:1}}" role="menuitem">item 1</div><div class="menu-item menu-item_theme_islands menu-item_disabled i-bem" data-bem="{&quot;menu-item&quot;:{&quot;val&quot;:2}}" role="menuitem">item 2</div></div>
```
narqo: There is no `id` attributes in `menu-item` – is this ok?

common.blocks/menu/menu.bh.js
```diff
@@ -31,6 +37,8 @@ module.exports = function(bh) {
                     var i = 0, itemOrGroup;
                     while(itemOrGroup = content[i++]) {
                         if(itemOrGroup.block === 'menu-item') {
+                            itemOrGroup.attrs = (itemOrGroup.attrs || {});
+                            itemOrGroup.attrs.id = itemOrGroup.attrs.id || ctx.generateId(); // Support a11y
```
narqo: AFAI: `ctx.generateId()` returns uniq id, for every context `ctx`. If so, all menu items would have the id

//cc @mishanga

qfox: @narqo Each call uses precalculated `_expandoId` (date), stored in BH, and numeric `lastGenId` in static scope. It means no. It returns unique id for each BH instance and each `generateId` call.

https://github.com/bem/bh/blob/master/lib/bh.js#L245

common.blocks/menu/menu.js
qfox: :+1: for caching.

common.blocks/menu/menu.bemhtml
qfox: :+1: Nice catch

common.blocks/menu/menu.bh.js
narqo: Oh, sh.. But this means that `generateId()` [from BEMHTML](https://github.com/bem/bem-core/blob/6ce978f653cd84a7a490e654d6c6dbf378cb7546/common.blocks/i-bem/i-bem.bemhtml#L241-L243) and bh works in different ways?

It looks like, `bh`'s `ctx.generateId()` is `this.identify()` from BEMHTML :(

qfox: @narqo :+1: Looks like.

common.blocks/menu/menu.bemhtml
hcodes: Да.

common.blocks/menu/menu.js
hcodes: Экономия на спичках.

Еще раз, id в menu-item нужен для связи выделенного пункта в menu, а именно атрибута aria-activedescendant. В один момент нужен только один id.
Здесь достаточно динамически добавлять уникальный id при выборе пункта и удалять при снятии.
Лучше отказаться от расстановки id в шаблонах и оставить всю логику в js на _onHoverItem.

common.blocks/menu/menu.bemhtml
hcodes, line 52: Видел, но тогда почему `tabIndex` в BEMJSON, а не `tabindex`?

common.blocks/menu-item/menu-item.js
hcodes, line 42: По codestyle есть только jscs, остальные случаи нужно описать.

common.blocks/menu/menu.js
narqo: > Здесь достаточно динамически добавлять уникальный id при выборе пункта и удалять при снятии.
> Лучше отказать от расстановки id в шаблонах и оставить всю логику в js на _onHoverItem.

I've spent some time thinking about this, and how I think this could work better. And even more: it seems, that you don't need to generate uniq id on every `hovered` call, as there can't be two `menu` (or `select` / `suggests`) with hovered items. All you  need is `identify(this)`

common.blocks/menu/menu.js
```diff
@@ -134,8 +134,10 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
...
             this._scrollToItem(this._hoveredItem = item);
         } else if(this._hoveredItem === item) {
+            this.domElem.removeAttr('aria-activedescendant');
```
narqo, line 147: Why do we need this?

common.blocks/menu-item/menu-item.js
narqo, line 42: It's weird to write code in the different manner within one file. Don't you think that code, that is written already, and that is in front of your face now, is much clearer describes the code-style than some sort of documentation somewhere?

common.blocks/menu/menu.bemhtml
```diff
@@ -26,7 +26,7 @@ block('menu')(
                             iterateItems(itemOrGroup.content);
                         }
                     }
-                };
+                }).bind(this);
```
aristov, line 29: Этот `bind` уже не нужен.

PR #1383, opened 2015-02-28
menu: two _onKeyDown handler binds if block is inited on focus (close #1381)

### LINKED ISSUE #1381: menu: two _onKeyDown handler binds if block inited on focus
https://github.com/bem/bem-components/blob/v2/common.blocks/menu/menu.js#L34
https://github.com/bem/bem-components/blob/v2/common.blocks/menu/menu.js#L42
If `menu` block inited with tab focus `_onKeyDown` handler bound twice. It causes scrolling two items with one cursor keypress.

dfilatov:
cc @narqo @Guria

### DIFF
--- a/common.blocks/menu/menu.js
+++ b/common.blocks/menu/menu.js
@@ -30,25 +30,6 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
                     index : 0,
                     time : 0
                 };
-
-                this.hasMod('focused') && this.bindToDoc('keydown', this._onKeyDown);
-            }
-        },
-
-        'focused' : {
-            'true' : function() {
-                this.__base.apply(this, arguments);
-                this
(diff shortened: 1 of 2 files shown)


Guria:
lgfm

common.blocks/menu/menu.js
```diff
@@ -131,6 +112,23 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
         return null;
     },
 
+    /** @override **/
+    _onFocus : function() {
```
narqo, line 116: Does this handler still work if block had had focus before it was inited?

narqo:
The code looks good. If there are no problems with initial focus – :snowflake:

common.blocks/menu/menu.js
dfilatov, line 116: Good catch :+1: Seems that no.
I think I've fixed it by the second commit. Take a look.

qfox:
:+1:

Issue #1390, opened 2015-03-05
Remove highlight control when tap on touch devices


hcodes:

hcodes:
Closes #1391

narqo:
Fixed in #1391

PR #1391, opened 2015-03-06
 Remove highlight control when tap on touch devices #1390

hcodes:


design/common.blocks/link/_theme/link_theme_islands.styl
```diff
@@ -47,4 +47,9 @@
             font-size: 18px;
         }
     }
+
+    &.link_disabled
+    {
+        -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
```
sipayRT, line 53: can't find any highlight on tap for disabled link

sipayRT, line 53: sry, already found

sipayRT:
maybe it will be better to move these changes to the `touch` level?

tadatuta:
> maybe it will be better to move these changes to the touch level?

we declare that desktop sites should work fine on touch devices

sipayRT:
ok, tnx.
@hcodes  works fine for me

PR #1393, opened 2015-03-06
dropdown: get rid of wrapping root `div` node (close #1392)

### LINKED ISSUE #1392: dropdown: get rid of wrapping root `div` node
Now `dropdown` has wrapping `div` that changes display mode of it's switcher (e.g. `dropdown_switcher_link` becomes `display: block` although it's based on `link` block, that is `display: inline`). I propose to get rid of this root node, make `dropdown` as mix to switcher and popup and link it by `id` js-parameter, so we can use `dropdown` as inline element without need of writing additional styles for it. To be more exact, `dropdown` will inherit display mode of it's switcher after this change.

aristov:
Resolves #1392


tadatuta:
why not to generate `id` automatically by default?

aristov:
@tadatuta I generate `id` automatically in template, but if I don't set `id` manually in tmpl-specs, then I get difference in `data-bem` attribute, because templaters generate a new `id` each time.

tadatuta:
:+1:

aristov:
@tadatuta maybe we can consider this case in `html-differ` and ignore difference for `id` param in `data-bem` attribute.

tadatuta:
@aristov not sure if it's possible as we need to check that `id` is the same for two corresponding DOM nodes.

narqo:
1) use `elem` instead of custom `mode`s in BEMHTML as it was before
2) add tests for proper destruct of `popup`

common.blocks/dropdown/dropdown.bemhtml
```diff
@@ -1,8 +1,19 @@
 block('dropdown')(
-    js()(true),
+    def()(function() {
+        this.ctx.js = this.extend({ id : this.generateId() }, this.ctx.js);
+        applyCtx([apply('switcher'), apply('popup')]);
```
narqo:
``` js
applyCtx([
  {
    elem : 'switcher',
    js : this.extend(apply('js'), this.ctx.js),
    ···
  },
  { elem : 'popup' }
])
```

common.blocks/dropdown/_switcher/dropdown_switcher_link.js
```diff
@@ -11,7 +11,7 @@ modules.define('dropdown', ['link'], function(provide, _, Dropdown) {
  */
 provide(Dropdown.decl({ modName : 'switcher', modVal : 'link' }, null, /** @lends dropdown */{
     live : function() {
-        this.liveInitOnBlockInsideEvent('click', 'link', this.onSwitcherClick);
+        this.liveInitOnBlockEvent('click', 'link', this.onSwitcherClick);
```
narqo, line 14: Think about moving `onSwitcherClick` to prototype.

narqo, line 14: This part is not that obvious. As far as I remember it was made this way not but the mistake. One must check how clear this API will be from the pov of block inheritance. E.g. what will happen if someone will redefine `onSwitcherClick` within modifier without calling base?

aristov, line 14: cc @dfilatov

aristov, line 14: I propose to think about it separately from this issue.

deeonis, line 14: cc @dfilatov

aristov, line 14: Considered to do #1502 in `v3`.

aristov:
TODO: ensure that dropdowns mixes work

aristov:
TODO: ensure that `popup`'s and `dropdown__switcher`'s mixes work

aristov:
For me: add `return` before `apply*`, squash and merge.

Issue #1394, opened 2015-03-08
Attach: describe for name BEMJSON API field


alexbaumgertner:
There is no information about `name` field both in 
https://github.com/bem/bem-components/blob/v2/common.blocks/attach/attach.en.md
https://github.com/bem/bem-components/blob/v2/common.blocks/attach/attach.ru.md

Here the code:
https://github.com/bem/bem-components/blob/v2/common.blocks/attach/__control/attach__control.bemhtml#L11

So you can set `name` for data, which will be upload by `attach` block.

PR #1397, opened 2015-03-11
attach: `name` field description added

innabelaya:
closes https://github.com/bem/bem-components/issues/1394


alexbaumgertner:
Thanks a lot!

PR #1410, opened 2015-03-17
Get rid of symlinks for better Windows support

tadatuta:
#1200

### DIFF
--- a/common.blocks/popup/popup.tests/gemini.blocks/test-nested/test.styl
+++ b/common.blocks/popup/popup.tests/gemini.blocks/test-nested/test.styl
@@ -1,1 +1,1 @@
-../../simple.blocks/test/test.styl
+@import "../../../../../common.blocks/popup/popup.tests/simple.blocks/test/test.styl";
--- a/common.blocks/popup/popup.tests/gemini.blocks/test-wrap/test-wrap.styl
+++ b/common.blocks/popup/popup.tests/gemini.blocks/test-wrap/test-wrap.styl
@@ -1,1 +1,2 @@
-../../simple.blocks/test-wrap/test-wrap.styl
+@import "../../../../../common.blocks/popup/popup.tests/simple.blocks/test-wrap/test-wrap.styl";
+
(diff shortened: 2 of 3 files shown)


tadatuta:
// cc @andrewblond it looks rather ugly :(

common.blocks/popup/popup.tests/gemini.blocks/test-nested/test.styl
```diff
@@ -1 +0,0 @@
-../../simple.blocks/test/test.styl
```
narqo, line 1: Why not to use borschik here?

tadatuta, line 1: I have no idea how to use it here :(

// cc @andrewblond

blond, line 1: `stylus` and `enb-stylus` not support `borschik`, and will not be.

narqo, line 1: @andrewblond but how is freeze work now? What are the problems to use borschik for such cases?

qfox, line 1: Probably because of borschik (as post-processor) work will happen after pre-processor?

narqo, line 1: So stylus resolves `url`s for background-image properly during the build, but it doesn't do so in case of `@import`s, am I wrong? In other case, I don't understand why there is `../../../../../` in the path

aristov, line 1: The last question is still relevant.

blond, line 1: This "feature" of the `enb-bem-examples`: create symlinks, but does not change contents of files.

aristov, line 1: @blond can we decide this "feature" is a bug and fix it?

blond, line 1: It is very difficult. It will be necessary to rewrite all the code around the `enb-bem-examples`.

tadatuta, line 1: created issue https://github.com/enb-bem/enb-bem-examples/issues/14

tadatuta:
TODO: add comments and merge

aristov:
@tadatuta ping

PR #1483, opened 2015-04-24
checkbox: Doesn't work on touch devices

### LINKED ISSUE #1472: checkbox: Doesn't work on touch devices
`checkbox` doesn't work properly on touch devices, at least on iPad/iPhone under the iOS 8.3 as well as touch devices emulator in Chrome.

In the [checkbox](https://en.bem.info/__example/bem-components/v2.1.0/desktop.examples/checkbox/5nmu9H4NkVi7P48W_Yx7tHZy6Hk/5nmu9H4NkVi7P48W_Yx7tHZy6Hk.html) example it is impossible to toggle `checked` state by tapping on the checkbox itself. While tapping on the `label` text works as expected.

See ftlabs/fastclick#351 for additional details

narqo:
fix #1472

See ftlabs/fastclick#351 for reference

/cc @tadatuta @mishanga can't this changes bring some new sideeffects?

### DIFF
--- a/gemini/checkbox.gemini.js
+++ b/gemini/checkbox.gemini.js
@@ -23,7 +23,7 @@ gemini.suite('checkbox', function(root) {
                     element
                 ];
             } else {
-                element = checkboxEnabledSelector + ' .checkbox__control';
+                element = checkboxEnabledSelector + ' .checkbox';
                 captureElements = [checkboxEnabledSelector];
             }
 
@@ -37,8 +37,10 @@ gemini.suite('checkbox', function(root) {
                     .capture('hovered', function(actions) {
                         actions.mouseMove(this.checkbox);
                     })
-                    .capture('focused-hard', function(actions) {
-                        actions.focus(this.checkbox);
+                    .capture('focused-hard', function(actions, find) {
+                        // NOTE: can't focus on `<label class="checkbox">`, so find `.checkbox__control` explicitly
+                        // see https://github.com/gemini-testing/gemini/issues/172
+                        actions.focus(find(element + '__control'));
                     })
                     .capture('focused-checked', function(actions) {
                         actions.click(this.checkbox);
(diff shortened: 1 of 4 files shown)


narqo:
It feels like the current problem is with our gemini tests for `checkbox`. [We're manipulating `.checkbox__control` explicitly](https://github.com/bem/bem-components/blob/2e8444a4ad7bf3a859748f377749d9fb65541ae1/gemini/checkbox.gemini.js#L26), but it feels like we need to use `.checkbox` as it looks like more fair way to iterate with the block.

I've checked test page with checkbox in Firefox (v35) and everything looks fine there. @tadatuta @sipayRT can you help me here?

sipayRT:
I'll check it

narqo:
@sipayRT any luck here?

sipayRT:
@narqo yes, I checked run tests on `.checkbox` instead of `.checkbox__control` - tests was written correctly because we can't get focused state on `<label>` tag. So `.checkbox__control` should be focused and `.checkbox` should be captured. I'm trying to fix gemini-errors now (but tests are running to slow now :( )

aristov:
@narqo ping

gemini/checkbox.gemini.js
```diff
@@ -37,8 +37,9 @@ gemini.suite('checkbox', function(root) {
                     .capture('hovered', function(actions) {
                         actions.mouseMove(this.checkbox);
                     })
-                    .capture('focused-hard', function(actions) {
-                        actions.focus(this.checkbox);
+                    .capture('focused-hard', function(actions, find) {
+                        // NOTE: can't focus on `<label>`, so find `checkbox__control` explicitly
+                        actions.focus(find(element + '__control'));
```
narqo, line 43: @sipayRT @tadatuta i've fixed `gemini` tests to use `checkbox` instead of `checkbox__control`. The only problem is with with `focus()` element. But I'm sure, it's a gemini bug (gemini-testing/gemini#172)

narqo:
Finally, I've fixed gemini tests! But we still has a massive yastatic.net + hsts problems in all PRs.

/cc @tadatuta @sipayRT

PR #1487, opened 2015-04-27
select: doesn't open by "space" at first time (close #1486)

### LINKED ISSUE #1486: select: doesn't open by "space" at first time

dfilatov:
cc @narqo @veged

### DIFF
--- a/common.blocks/control/control.js
+++ b/common.blocks/control/control.js
@@ -106,7 +106,9 @@ provide(BEMDOM.decl(this.name, /** @lends control.prototype */{
 }, /** @lends control */{
     live : function() {
         this
-            .liveBindTo('control', 'focusin', this.prototype._onFocus)
+            .liveBindTo('control', 'focusin', function() {
+                this._focused || this._onFocus(); // to prevent double call of _onFocus in case of init by focus
+            })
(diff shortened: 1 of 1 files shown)


common.blocks/control/control.js
```diff
@@ -106,7 +106,9 @@ provide(BEMDOM.decl(this.name, /** @lends control.prototype */{
 }, /** @lends control */{
     live : function() {
         this
-            .liveBindTo('control', 'focusin', this.prototype._onFocus)
+            .liveBindTo('control', 'focusin', function() {
```
narqo, line 109: Why do we have [separate methods in the prototype](https://github.com/bem/bem-components/blob/2e8444a4ad7bf3a859748f377749d9fb65541ae1/common.blocks/select/select.js#L221-L231) in all cases expect this one? May be we can add `_onRequestFocus` or `_onFocusin` here just to stay neat?

dfilatov, line 109: Because I don't want to increase a mess with managing focus state which we've already had.

narqo, line 109: Ok

narqo:
:fire:

PR #1493, opened 2015-04-30
Added srcset attribute support

f0rmat1k:
Currently, in common case `src-set` is a good solution for retina support. It is supported in the latest versions of browsers, so polyfill not required. Yes, we can use just `attrs` for it, but It is not very convenient.

If you find this idea good, i'l add to this pull request bemhtml implementation, readme, support for the attribute `sizes` and add tests.


tadatuta:
browser support is still quite weak :(
and we should guarantee all provided api to work fine on all supported browsers so it's a bit premature. at least without polyfills.

f0rmat1k:
Ok

PR #1495, opened 2015-05-05
common.blocks: prepare for bem-xjst@2.x

indutny:
- Remove `match()` with inline value
- Always return when doing `applyNext()`/`applyCtx()`

cc @veged

### DIFF
--- a/common.blocks/menu-item/_type/menu-item_type_link.bemhtml
+++ b/common.blocks/menu-item/_type/menu-item_type_link.bemhtml
@@ -1,9 +1,13 @@
-block('menu-item').mod('type', 'link').mod('disabled', true).match(!this._menuItemDisabled).def()(function() {
-    apply({ _menuItemDisabled : true });
+block('menu-item').mod('type', 'link').mod('disabled', true).match(function() {
+    return !this._menuItemDisabled;
+}).def()(function() {
+    return applyNext({ _menuItemDisabled : true });
 });
 
-block('link').match(this._menuItemDisabled).def()(function() {
+block('link').match(function() {
+    return this._menuItemDisabled;
+}).def()(function() {
     delete this._menuItemDisabled;
     this.mods.disabled = true;
-    apply();
+    return applyNext();
 });
--- a/common.blocks/select/_mode/select_mode_radio.bemhtml
+++ b/common.blocks/select/_mode/select_mode_radio.bemhtml
@@ -1,7 +1,5 @@
 block('select').mod('mode', 'radio')(
-    def().match(function() {
-        return this._checkedOptions;
-    })(function() {
+    def().match(function() { return this._checkedOptions; })(function() {
         var checkedOptions = this._checkedOptions,
             firstOption = this._firstOption;
         if(firstOption && !checkedOptions.length) {
--- a/common.blocks/select/_mode/select_mode_radio.bemhtml
+++ b/common.blocks/select/_mode/select_mode_radio.bemhtml
@@ -1,12 +1,14 @@
 block('select').mod('mode', 'radio')(
-    def().match(this._checkedOptions)(function() {
+    def().match(function() {
+        return this._checkedOptions;
+    })(function() {
         var checkedOptions = this._checkedOptions,
             firstOption = this._firstOption;
         if(firstOption && !checkedOptions.length) {
             firstOption.checked = true;
             checkedOptions = [firstOption];
         }
-        applyNext({ _checkedOption : checkedOptions[0] });
+        return applyNext({ _checkedOption : checkedOptions[0] });
     }),
 
     content()(function() {
--- a/common.blocks/select/select.tests/gemini.blocks/select/select.bemhtml
+++ b/common.blocks/select/select.tests/gemini.blocks/select/select.bemhtml
@@ -7,5 +7,5 @@ block('popup').match(function() { return this._selectCls; }).def()(function() {
     // with special classes, so gemini tests could find them
     this.ctx.cls = this._selectCls + '-popup';
     delete this._selectCls;
-    return apply();
+    return applyNext();
 });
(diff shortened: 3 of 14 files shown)


common.blocks/menu-item/_type/menu-item_type_link.bemhtml
```diff
@@ -1,9 +1,13 @@
...
 });
 
-block('link').match(this._menuItemDisabled).def()(function() {
+block('link').match(function() {
+    return this._menuItemDisabled;
+}).def()(function() {
     delete this._menuItemDisabled;
     this.mods.disabled = true;
-    apply();
+    return applyNext();
```
Guria, line 12: здесь точно надо apply менять на applyNext?

veged, line 12: @Guria для новой версии bem-xjst нужны такие изменения — по семантике это вполне корректно, просто раньше мы "экономили" `applyNext` т.к. был кастомный гард про `_menuItemDisabled`

qfox, line 12: @veged Но гард все еще оставляем?

veged, line 12: @zxqfox да, т.к. он нужен по смыслу

common.blocks/select/select.tests/gemini.blocks/select/select.bemhtml
```diff
@@ -1,11 +1,13 @@
...
-block('popup').match(this._selectCls).def()(function() {
+block('popup').match(function() {
+    return this._selectCls;
+}).def()(function() {
     // NOTE: we "mark" popups which are inside select block
     // with special classes, so gemini tests could find them
     this.ctx.cls = this._selectCls + '-popup';
     delete this._selectCls;
-    apply();
+    return apply();
```
Guria: @veged а здесь?

veged: хм... вообще хороший вопрос — @indutny если тут можно `apply`, то и выше тоже можно `apply`

indutny: Ну у нас `apply()` теперь нету, есть только `apply('mode')` и `applyNext()`. `applyNext()` здесь подходит по семантике.

veged: тогда тут тоже нужно поправить на `applyNext`

indutny: Аааа, точно. Я не заметил.

indutny: Fixed.

qfox: Слушайте, а на это не надо тесты? Или как оно работало без applyNext?

common.blocks/select/_mode/select_mode_radio.bemhtml
```diff
@@ -1,12 +1,12 @@
...
             firstOption = this._firstOption;
         if(firstOption && !checkedOptions.length) {
             firstOption.checked = true;
             checkedOptions = [firstOption];
         }
-        applyNext({ _checkedOption : checkedOptions[0] });
+        return applyNext({ _checkedOption : checkedOptions[0] });
     }),
 
     content()(function() {
```
tadatuta, line 12: when compiled with `bem-xjst@2.x`, `this._checkedOption` is undefined in `content()()` for some reason
// cc @indutny

indutny, line 12: How did you test it? Could it be that `_checkedOptions` was `undefined` too?

tadatuta, line 12: looks like it is defined at https://github.com/bem/bem-components/blob/feature/prepare-for-bem-xjst-v2/common.blocks/select/select.bemhtml#L34 and undefined in `block('select').mod('mode', 'radio')`

indutny, line 12: So it was present in select.bemhtml? It is especially interesting to see, because this block appears to be present in the one of the desktop.pages, and I have verified that generated html is the same with old and new bem-xjst.

tadatuta, line 12: please take a look at https://github.com/tadatuta/select-template-issue-demo
not the very best test case but hope it will help

indutny, line 12: @tadatuta very interesting!

@veged the situation is following: there are `block('select').mod('mode', 'radio')` which has higher priority than `block('select').def()` in the `.bemhtml` file. When `block('select').def()` calls `applyNext()` - it is only invoking the matches in the blocks with lower priority than `block('select').def()`, skipping the `block('select').mod('mode', 'radio')`.

Looks like our idea about incrementing index and invoking next templates is not as **viable** as we thought it will be. What do you think?

indutny, line 12: @tadatuta may I ask you to give a try to the latest bem-xjst 2.x master? It might be fixed by https://github.com/bem/bem-xjst/commit/ee7d1defa4d1c2c7fb3406e8c8a7afe4c0762d04

tadatuta, line 12: yes, this issue is resolved, thanks!

aristov:
Is it WIP?

narqo:
@indutny ping!

indutny:
@narqo I think all issues are resolved now. Feel free to land it if it LGTY

tadatuta:
I think we should merge https://github.com/bem/bem-core/pull/1022, release a version and then update dependency. Otherwise there's just no reason.

indutny:
It doesn't seem like it is blocking this one, or am I missing something?

tadatuta:
it's not blocking but it's kinda useless without support in `bem-core`

indutny:
@tadatuta do we want to merge this, considering that bem/bem-core#1022 was merged?

tadatuta:
yeah. but looks like rebase is needed.

aristov:
@indutny rebase and merge, please.

Issue #1516, opened 2015-05-27
[menu] global context properties


Guria:
`_firstItem` and `_checkedItems` properties [set as global](https://github.com/bem/bem-components/blob/v2.1.1/common.blocks/menu/menu.bemhtml#L36-L37) in context.

``` js
          { block:'menu', content:[{block:'menu-item', content:'asdasd'}]},
          { block: 'debug'},
```

``` js
block('debug').content()(function(){
    return JSON.stringify(this._firstItem);
});
```

``` html
<div class="menu menu__control i-bem" data-bem="{&quot;menu&quot;:{}}" role="menu" tabindex="0">
  <div class="menu-item i-bem" data-bem="{&quot;menu-item&quot;:{}}" role="menuitem">asdasd</div></div>
<div class="debug">{"block":"menu-item","content":"asdasd","mods":{}}</div>
```

narqo:
Nice catch, thanks! We will fix it in the nearest future

narqo:
Actually, it shouldn't hurt any "real life" usage (as these flags will be reassigned for each menu instance). 

But anyway, could you make a separate PR for the issue, so we will merge it?

Guria:
Okay

qfox:
:+1: for that fix but looks like it's breaking change.

Guria:
@zxqfox #1523 just bug fix, #1520 more serious change but I don't see what it can brake

PR #1520, opened 2015-05-28
[menu] move specific template logic which deal with menu-item's checked mod from menu to common menu_mode

### LINKED ISSUE #1513: [menu] base template for block contains logic only for one mod.
https://github.com/bem/bem-components/commit/da75b8177b9aa73ab80e068a7d55727c6daf6803 introduced logic which demanded only by `menu_mode_radio` template.
It iterates over content to set properties `_firstItem` and `_checkedItems`. Iteration function has assumption that `menu` should contain only `menu-item` or `menu__group`. It very inconvinient when we try to put inside block `menu` another elements or blocks which will be replaced with `menu-item` in later processing.

I suppose two enhancements:
- Move that specific code under `mod('mode', 'radio')` matcher.
- Make content iteration function indiferrent to block's content:
  - add `ctx.content = [].concat(ctx.content)` instead of `if(!this.isArray(ctx.content)) throw Error`
  - check on content property existance (`else if(ctx.content) { // menu__group`) before recursive call

Additionally we can support case of resolving blocks to `menu-item` which cannot be catched by iteration function and add js behavior in init section of `menu_mode_radio`, and document that in such cases live initialization should be disabled. But I think it can and should be done at project level.

/cc @hcodes @tadatuta

Guria:
iterate through content only within all menu_mode_*
do it in `content` mode instead of `def`
use applyNext instead of setting global ctx properties

resolves #1513  #1516

### DIFF
--- a/common.blocks/menu/_mode/menu_mode.bemhtml
+++ b/common.blocks/menu/_mode/menu_mode.bemhtml
@@ -1,0 +1,43 @@
+block('menu')
+    .match(function(){
+        return typeof this.mods.mode !== 'undefined';
+    })
+    .content()(function() {
+        var ctx = this.ctx,
+            mods = this.mods,
+            firstItem,
+            checkedItems = [];
+
+        if(ctx.content) {
+            var isValDef = typeof ctx.val !== 'undefined',
+                containsVal = function(val) {
+                    return isValDef &&
+                        (mods.mode === 'check'?
+                            ctx.val.indexOf(val) > -1 :
+                            ctx.val === val);
+                },
+                iterateItems = function(content) {
+                    var i = 0, itemOrGroup;
+                    while(itemOrGroup = content[i++]) {
+                        if(itemOrGroup.block === 'menu-item') {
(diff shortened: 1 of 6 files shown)


Guria:
Why js tests could fail if all templates tests are ok? JS wasn't touched in commit.

Guria:
It seems that spec tests doesn't include menu_mode.bemhtml into test bundle while tmpl-specs works as expected.

common.blocks/menu/_mode/menu_mode.bemhtml
```diff
@@ -0,0 +1,40 @@
+block('menu')
+    .match(function(){
+        return this.ctx.mods['mode'];
```
tadatuta: I think it should be like this:

``` js
return typeof this.mods.mode !== 'undefined';
```

Guria:
cc @tadatuta @blond please help
I have no enough experience to understand why menu_mode doesn't included in specs tests

Guria:
@tadatuta I have resolved issue with specs tests.
Hope this pr will be merged

PR #1523, opened 2015-05-28
menu: set properties within applyNext()

### LINKED ISSUE #1516: [menu] global context properties

Guria:
use applyNext instead of global ctx properties
resolves #1516

### DIFF
--- a/common.blocks/menu/menu.bemhtml
+++ b/common.blocks/menu/menu.bemhtml
@@ -33,10 +33,9 @@ block('menu')(
             iterateItems(ctx.content);
         }
 
-        this._firstItem = firstItem;
-        this._checkedItems = checkedItems;
-
         applyNext({
+            _firstItem : firstItem,
+            _checkedItems : checkedItems,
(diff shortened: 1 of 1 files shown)


narqo:
Looks good, thanks

PR #1526, opened 2015-05-29
Load jquery from cdnjs.cloudflare.com instead of yastatic.net

narqo:
Trying to fix yastatic.net + CORS / HSTS problems in the current phantomjs runtime.

Currently all our specs fail with errors like the one below:

```
Failed to start mocha: Init timeout
Error loading resource http://yastatic.net/jquery/2.1.3/jquery.min.js (5). Details: Operation canceled
Unsafe JavaScript attempt to access frame with URL about:blank from frame with URL file:///home/travis/build/bem/bem-components/node_modules/enb-bem-specs/node_modules/mocha-phantomjs/lib/mocha-phantomjs.coffee. Domains, protocols and ports must match.
```

It seems that HSTS headers are the reason. But it's not clear, how should it be fixed as there are several bugs in the phantomjs related projects on the github, and non of the solution (e.g. use https-only or downgrade phantomjs to 1.9.7) seems doesn't work

### DIFF
--- a/.enb/make.js
+++ b/.enb/make.js
@@ -377,7 +377,8 @@ function getTestLevels(platform) {
 function getSpecLevels(platform) {
     return [].concat(
         { path : path.join('libs', 'bem-pr', 'spec.blocks'), check : false },
-        getSourceLevels(platform)
+        getSourceLevels(platform),
+        'specs.blocks'
     );
 }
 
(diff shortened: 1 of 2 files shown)


tadatuta:
@narqo have you tried to hardcode `https` protocol?

narqo:
yep, see #1524

narqo:
ping!

tadatuta:
lgtm

narqo:
Ok, I've issued #1528 for future investigations. Hope someone will try to fix it

PR #1533, opened 2015-06-02
menu: check content propertie existence

### LINKED ISSUE #1513: [menu] base template for block contains logic only for one mod.

Guria:
partially resolves #1513

### DIFF
--- a/common.blocks/menu/menu.bemhtml
+++ b/common.blocks/menu/menu.bemhtml
@@ -22,7 +22,7 @@ block('menu')(
                                 (itemOrGroup.mods = itemOrGroup.mods || {}).checked = true;
                                 checkedItems.push(itemOrGroup);
                             }
-                        } else { // menu__group
+                        } else if(itemOrGroup.content) { // menu__group
                             iterateItems(itemOrGroup.content);
                         }
                     }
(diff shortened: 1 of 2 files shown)


tadatuta:
@narqo lgtm (gemini failed because of external account), should I merge and add it to 2.2.0?

narqo:
ok

PR #1567, opened 2015-07-01
button: fix wrong definition of real button

### LINKED ISSUE #1566: button: wrong definition of real button
Now in templates we check just `type` modifier:

```
var modType = ctx.mod('type'),
      isRealButton = !modType || modType === 'submit';
```

But if we will change in BEMJSON tag, for example, to `<div>`, variable `isRealButton` still will be `true`. It's wrong because we can't add some attributes such as `disabled` to `<div>`. 

In the `attach` block it's very important because we use `<span>` tag instead of `<button>`. As a result — disabled button in `attach` block will be disabled only visually for a11y's programs.

sipayRT:
Closes #1566


sipayRT:
@tadatuta @narqo need review

tadatuta:
lgtm

common.blocks/button/button.bemhtml
```diff
@@ -1,4 +1,11 @@
 block('button')(
+    def()(function() {
+        var isRealButton = (['button', 'input'].indexOf(apply('tag')) >= 0) &&
```
narqo: We have a `NOTE` comment in the `content` mode below about "mode call optimisations". But here we call `tag` which will be called again within `applyNext()`. Maybe we should be more strict?

tadatuta: any ideas on how to solve the issue without it?

sipayRT: @narqo I can replace it with `this.ctx.tag || 'button'`. Is it suits you?

sipayRT: @narqo ping

narqo: 1) What if we move the content of `tag` mode inside `def`?

``` js
def()(() => {
  this.ctx.tag || (this.ctx.tag = 'button');
  // ···
})
```

2) I think we can omit `input` case check and leave only `tag === 'button'`. You can't use `input` as a tag of the `button`, as it must have `button__text` and `button__icon` in it's content and this is unachievable.

sipayRT: 1) in this case you should change `def` mode(not `tag`) in attach too. And it's very strange to change tag in `def` mode.
2) agree

sipayRT: @narqo ping

narqo: > in this case you should change def mode(not tag) in attach too.

I don't understand why you have to change anything in `attach`. It [passes button's tag explicitly as a bemjson](https://github.com/bem/bem-components/blob/v2/common.blocks/attach/attach.bemhtml#L13-L17).

> And it's very strange to change tag in def mode

Yes, as well as to call `apply('tag')` twice. At least, my variant should be more performant, I guess.

> agree

so?

sipayRT: @narqo 
For example, we have an attach with such BEMJSON as:

```
    {
        block : 'attach',
        name : 'attach-name',
        button : { block : 'button', mods : { theme : 'islands', size : 'm' } },
        noFileText : 'no file'
    }
```

In this case in `def`-mode:

```
this.ctx.tag || (this.ctx.tag = 'button');
var isRealButton = (['button'].indexOf(this.ctx.tag) >= 0) &&
                (!this.mods.type || this.mods.type === 'submit');
```

So, `isRealButton` will be `true` and `type` attribute will be set as `'button'` (https://github.com/bem/bem-components/blob/issues/1566%40v2/common.blocks/button/button.bemhtml#L39).
After it in [attach__button in `tag` mode](https://github.com/bem/bem-components/blob/issues/1566%40v2/common.blocks/attach/__button/attach__button.bemhtml#L3) we will change tag to `<span>`. As a result, it will be `<span type="button">`. And to fix it now we should add `def` mode to `attach__button` too.

narqo: Ok, now I get it. 

But at least, please, let's change this ridiculous `['button'].indexOf(tag) >= 0` to `tag === 'button'`

sipayRT: of course. Without `'input'` it has no sense

sipayRT:
@narqo need your final review

common.blocks/button/button.bh.js
```diff
@@ -1,12 +1,14 @@
 module.exports = function(bh) {
 
     bh.match('button', function(ctx, json) {
+        ctx.tag(json.tag || 'button'); // NOTE: need to predefine tag
+
         var modType = ctx.mod('type'),
-            isRealButton = !modType || modType === 'submit';
+            isRealButton = (['button', 'input'].indexOf(ctx.tag()) >= 0) &&
```
narqo: I think this part needs to be fixed as well

sipayRT: sorry. forgot for it. fixed

narqo:
Ok, let it be so

PR #1572, opened 2015-07-02
progressbar: improve JAWS readability

### LINKED ISSUE #1511: progressbar: improve JAWS readability
JAWS should read
- [x] `progressbar` as "progressbar"
- [x] `val` field + "percent"

Check and fix divergencies.

sipayRT:
Closes #1511

### DIFF
--- a/common.blocks/progressbar/progressbar.bemhtml
+++ b/common.blocks/progressbar/progressbar.bemhtml
@@ -1,14 +1,24 @@
 block('progressbar')(
+    def()(function() {
+        applyNext({ _val : this.ctx.val || 0 });
+    }),
 
     js()(function(){
-        return { val : this.ctx.val || 0 };
+        return { val : this._val };
+    }),
+
+    attrs()(function() {
+        return {
+            role : 'progressbar',
+            'aria-valuenow' : this._val + '%' // NOTE: JAWS doesn't add 'percent' automatically
+        };
     }),
 
     content()(
         function() {
             return {
                 elem : 'bar',
-                attrs : { style : 'width:' + (this.ctx.val || 0) + '%' }
+                attrs : { style : 'width:' + this._val + '%' }
             };
         },
         match(function() { return typeof this.ctx.content !== 'undefined'; })(function() {
--- a/common.blocks/progressbar/progressbar.bh.js
+++ b/common.blocks/progressbar/progressbar.bh.js
@@ -5,6 +5,10 @@ module.exports = function(bh) {
 
         ctx
             .js({ val : val })
+            .attrs({
+                role : 'progressbar',
+                'aria-valuenow' : val + '%'  // NOTE: JAWS doesn't add 'percent' automatically
+            })
             .content({
                 elem : 'bar',
                 attrs : { style : 'width:' + val + '%' }
--- a/common.blocks/progressbar/progressbar.js
+++ b/common.blocks/progressbar/progressbar.js
@@ -24,7 +24,8 @@ provide(BEMDOM.decl(this.name, /** @lends progressbar.prototype */{
      * @returns {progressbar} this
      */
     setVal : function(val) {
-        this.elem('bar').css('width', (this._val = val) + '%');
+        this.domElem.attr('aria-valuenow', (this._val = val) + '%');
+        this.elem('bar').css('width', val + '%');
         return this;
     },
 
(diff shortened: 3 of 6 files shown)


common.blocks/progressbar/progressbar.bemhtml
```diff
@@ -6,9 +6,15 @@ block('progressbar')(
...
     content()(
         function() {
+            var val = (this.ctx.val || 0) + '%';
+
             return {
                 elem : 'bar',
-                attrs : { style : 'width:' + (this.ctx.val || 0) + '%' }
+                attrs : {
+                    style : 'width:' + val,
+                    role : 'progressbar',
```
aristov: `role` and `aria-valuenow` attributes should be added to the root node of the block.

common.blocks/progressbar/progressbar.js
```diff
@@ -24,7 +24,9 @@ provide(BEMDOM.decl(this.name, /** @lends progressbar.prototype */{
      * @returns {progressbar} this
      */
     setVal : function(val) {
-        this.elem('bar').css('width', (this._val = val) + '%');
+        this.elem('bar')
+            .css('width', (this._val = val) + '%')
+            .attr('aria-valuenow', this._val);
```
aristov: You lose additional `%` symbol here.

sipayRT: But specification says that value should be as a number. I don't add `aria-value(min/max)` so `aria-valuenow` will be as `%` by default.

common.blocks/progressbar/progressbar.bemhtml
sipayRT: For this we should change `style` and `aria-valuenow` attributes in different places. It will be strange.
For example:

```
<div class="progressbar" aria-valuenow="20">
    <div class="progressbar__bar" style="width:20%"></div>
</div>
```

Two DOM-nodes will be changed each second while some process will be going. If this is not so terrible, I can move these attributes

aristov: I think it's not so terrible. Let's do it.

common.blocks/progressbar/progressbar.js
aristov: You add `%` in templates and lose it here. This is inconsistency. In addition, JAWS doesn't add "percent" automatically, so we are constrained to deviate from specification.

sipayRT:
@aristov all a11y attributes was moved to the root node

common.blocks/progressbar/progressbar.bemhtml
```diff
@@ -1,14 +1,27 @@
 block('progressbar')(
+    def()(function() {
+        var val = this.ctx.val || 0;
```
aristov: No need to create variable here.

common.blocks/progressbar/progressbar.bemhtml
```diff
@@ -1,14 +1,27 @@
...
 
     js()(function(){
-        return { val : this.ctx.val || 0 };
+        return { val : this._val };
+    }),
+
+    attrs()(function() {
+        return {
+            role : 'progressbar',
+            'aria-valuenow' : this._val + '%'
```
aristov: Add comment why we deviate from specification here.

sipayRT:
@aristov updated

common.blocks/progressbar/progressbar.bh.js
```diff
@@ -5,9 +5,15 @@ module.exports = function(bh) {
...
             .js({ val : val })
+            .attrs({
+                role : 'progressbar',
+                'aria-valuenow' : val + '%'  // NOTE: JAWS doesn't add 'percent' automatically
+            })
             .content({
                 elem : 'bar',
-                attrs : { style : 'width:' + val + '%' }
+                attrs : {
+                    style : 'width:' + val + '%'
```
aristov: Diff for no reason.

common.blocks/progressbar/progressbar.bemhtml
```diff
@@ -1,14 +1,26 @@
...
+            'aria-valuenow' : this._val + '%' // NOTE: JAWS doesn't add 'percent' automatically
+        };
     }),
 
     content()(
         function() {
             return {
                 elem : 'bar',
-                attrs : { style : 'width:' + (this.ctx.val || 0) + '%' }
+                attrs : {
```
aristov: Collapse code to single line.

sipayRT:
@aristov updated

aristov:
LGTM, wait for Gemini tests passing and merge.

PR #1575, opened 2015-07-07
attach: improve JAWS readability

### LINKED ISSUE #1491: attach: improve JAWS readability
JAWS should read
- [x] `attach` as "button"
- [x] `attach_disabled` as "button unavailable"
- [x] `button` field
- [x] `noFileText` field

Check and fix divergencies.

sipayRT:
Closes #1491

### DIFF
--- a/common.blocks/attach/attach.js
+++ b/common.blocks/attach/attach.js
@@ -80,8 +80,8 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends attac
             this.domElem,
             '<span class="' +
                 this.__self.buildClass('file') + '">' +
-                '<span class="' +
-                    this.__self.buildClass('text') + '">' +
+                '<span class="' + this.__self.buildClass('text') + '"' +
+                    ' aria-hidden="true">' +
                     escape.html(fileName) +
                 '</span>' +
                 '<i class="' + this.__self.buildClass('clear') + '"/>' +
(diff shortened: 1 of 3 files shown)


sipayRT:
@aristov this branch was rebased. need your review

common.blocks/attach/__no-file/attach__no-file.bh.js
```diff
@@ -1,5 +1,7 @@
 module.exports = function(bh) {
     bh.match('attach__no-file', function(ctx) {
-        ctx.tag('span');
+        ctx
+            .tag('span')
+            .attr('aria-hidden', true);
```
qfox: Incompatible with BH 4.1+, Use `'true'` instead of `true` ;-) upd: Or fix tests.

sipayRT: I'll do it in common a11y branch

aristov:
JAWS speaks native "browse" and "no file selected" regardless of `button` and `noFileText` fields. Try to hide native control from JAWS by using `role="presentation"` or `aria-hidden="true"`.

aristov:
`noFileText` field still not accessible.

sipayRT:
@aristov 

> Try to hide native control from JAWS

I tried to do it, but in case when some file was chosen, you can't prevent reading of this element (even `aria-hidden` didn't help). So I offer to disable elem `__text` to avoid double-reading of chosen filename.

PR #1578, opened 2015-07-10
menu, menu-item: improve JAWS readability

### LINKED ISSUE #1508: menu, menu-item: improve JAWS readability
JAWS should read
- [x] `menu_focused` as "menu"
- [x] menu activation (by spacebar) as the first `menu-item` of `menu_focused`
- [x] `menu_focused` modifier removal as "leaving menus"
- [x] `menu_disabled` as "menu unavailable"
- [x] `content` field of `menu` as list of `menu__group` and `menu-item` entities
- [x] `menu__group` with specified `title` field as "group start" / "group end"
- [x] `title` field of `menu__group`
- [x] `menu-item` as "menu"
- [x] `menu-item` within `menu` with specified `mode` modifier as "menu not checked"
- [x] `menu-item_checked` within `menu` with specified `mode` modifier as "menu checked"
- [x] `menu-item_disabled` as "menu unavailable"
- [x] `content` field of `menu-item`

Check and fix divergencies.

sipayRT:
Closes #1508 #1458

### DIFF
--- a/common.blocks/menu-item/menu-item.bemhtml
+++ b/common.blocks/menu-item/menu-item.bemhtml
@@ -8,5 +8,19 @@ block('menu-item')(
     js()(function() {
         return { val : this.ctx.val };
     }),
-    attrs()({ role : 'menuitem' })
+    attrs()(function(){
+        var mods = this.mods,
+            menuMode = this._menuMods && this._menuMods.mode,
+            role = menuMode?
+                        (menuMode === 'check'? 'menuitemcheckbox' : 'menuitemradio') :
+                        'menuitem',
+            attrs = {
+                role : role,
+                id : this.generateId(),
+                'aria-disabled' : mods.disabled,
+                'aria-checked' : menuMode && !!mods.checked
+            };
+
+        return attrs;
+    })
 );
--- a/common.blocks/menu-item/menu-item.bh.js
+++ b/common.blocks/menu-item/menu-item.bh.js
@@ -1,6 +1,10 @@
 module.exports = function(bh) {
     bh.match('menu-item', function(ctx, json) {
-        var menuMods = ctx.tParam('menuMods');
+        var menuMods = ctx.tParam('menuMods'),
+            menuMode = menuMods && menuMods.mode,
+            role = menuMode?
+                        (menuMode === 'check'? 'menuitemcheckbox' : 'menuitemradio') :
+                        'menuitem';
 
         menuMods && ctx.mods({
             theme : menuMods.theme,
@@ -9,6 +13,11 @@ module.exports = function(bh) {
 
         ctx
             .js({ val : json.val })
-            .attr('role', 'menuitem');
+            .attrs({
+                role : role,
+                id : ctx.generateId(),
+                'aria-disabled' : ctx.mod('disabled'),
+                'aria-checked' : menuMode && !!ctx.mod('checked')
+            });
     });
 };
--- a/common.blocks/menu-item/menu-item.deps.js
+++ b/common.blocks/menu-item/menu-item.deps.js
@@ -5,4 +5,10 @@
 {
     tech : 'spec.js',
     mustDeps : { tech : 'bemhtml', block : 'menu-item' }
+},
+{
+    tech : 'tmpl-spec.js',
+    mustDeps : [
+        { tech : 'bemhtml', block : 'link' }
+    ]
 }]
--- a/common.blocks/menu/menu.bemhtml
+++ b/common.blocks/menu/menu.bemhtml
@@ -38,13 +38,18 @@ block('menu')(
             _checkedItems : checkedItems,
             _menuMods : {
                 theme : mods.theme,
-                disabled : mods.disabled
+                disabled : mods.disabled,
+                mode : mods.mode
             }
         });
     }),
     attrs()(function() {
         var attrs = { role : 'menu' };
-        this.mods.disabled || (attrs.tabindex = 0);
+
+        this.mods.disabled?
+            attrs['aria-disabled'] = true :
+            attrs.tabindex = 0;
+
         return attrs;
     }),
     js()(true),
--- a/common.blocks/menu/menu.bh.js
+++ b/common.blocks/menu/menu.bh.js
@@ -1,18 +1,23 @@
 module.exports = function(bh) {
 
     bh.match('menu', function(ctx, json) {
-        var menuMods = {
-            theme : ctx.mod('theme'),
-            disabled : ctx.mod('disabled')
-        };
+        var mods = ctx.mods(),
+            menuMods = {
+                theme : mods.theme,
+                disabled : mods.disabled,
+                mode : mods.mode
+            },
+            attrs = { role : 'menu' };
 
         ctx
             .js(true)
             .tParam('menuMods', menuMods)
             .mix({ elem : 'control' });
 
-        var attrs = { role : 'menu' };
-        ctx.mod('disabled') || (attrs.tabindex = 0);
+        mods.disabled?
+            attrs['aria-disabled'] = true :
+            attrs.tabindex = 0;
+
         ctx.attrs(attrs);
 
         var firstItem,
--- a/common.blocks/menu/menu.js
+++ b/common.blocks/menu/menu.js
@@ -33,10 +33,27 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
             }
         },
 
-        'disabled' : function(modName, modVal) {
-            this.getItems().forEach(function(menuItem){
-                menuItem.setMod(modName, modVal);
-            });
+        'disabled' : {
+            '*' : function(modName, modVal) {
+                this.getItems().forEach(function(menuItem){
+                    menuItem.setMod(modName, modVal);
+                });
+            },
+            'true' : function() {
+                this.__base.apply(this, arguments);
+                this.domElem.attr('aria-disabled', true);
+            },
+            '' : function() {
+                this.__base.apply(this, arguments);
+                this.domElem.removeAttr('aria-disabled');
+            }
+        },
+
+        'focused' : {
+            '*' : function(_, modVal) {
+                this.__base.apply(this, arguments);
+                this._activedescendantControl(modVal);
+            }
         }
     },
 
@@ -135,9 +152,12 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
      */
     _onItemHover : function(item) {
         if(item.hasMod('hovered')) {
+            var itemId = item.domElem.attr('id');
             this._hoveredItem && this._hoveredItem.delMod('hovered');
+            this.domElem.attr('aria-activedescendant', itemId);
             this._scrollToItem(this._hoveredItem = item);
         } else if(this._hoveredItem === item) {
+            this.domElem.removeAttr('aria-activedescendant');
             this._hoveredItem = null;
         }
     },
@@ -217,6 +237,20 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
     _doesItemMatchText : function(item, text) {
         return !item.hasMod('disabled') &&
             item.getText().toLowerCase().search(text) === 0;
+    },
+
+    /**
+     * @param {Boolean} isFocused
+     * @private
+     */
+    _activedescendantControl : function(isFocused) {
+        var firstItemId = this.getItems()[0].domElem.attr('id');
+
+        if(isFocused) {
+            this.domElem.attr('aria-activedescendant', firstItemId);
+        } else {
+            this.domElem.removeAttr('aria-activedescendant');
+        }
     }
 }, /** @lends menu */{
     live : function() {
--- a/common.blocks/menu/menu.spec.js
+++ b/common.blocks/menu/menu.spec.js
@@ -90,6 +90,15 @@ describe('menu', function() {
                 menuItem.hasMod('disabled').should.be.false;
             });
         });
+
+        it('should set/remove "aria-disabled" attributes on disable', function() {
+            menu
+                .setMod('disabled')
+                .domElem.attr('aria-disabled').should.be.equal('true');
+
+            menu.delMod('disabled');
+            expect(menu.domElem.attr('aria-disabled')).to.be.undefined;
+        });
     });
 
     describe('hover on items', function() {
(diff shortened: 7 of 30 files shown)


sipayRT:
@aristov need your review

sipayRT:
@aristov ping. Need your review here

aristov:
2nd item of #1508 description is not satisfied.
When virtual cursor comes to `__group-title` DOM-node, JAWS reads "[title] group box", but shouldn't. Use [presentation role](http://www.w3.org/TR/wai-aria/roles#presentation) to avoid it.

sipayRT:
@aristov I tried to prevent it by `presentation role`, but it not works for me (I added this attribute - you can check it). To fix it we can add `'aria-hidden'` attribute for `group-title` element.

sipayRT:
@aristov tests are failed because of `role="presentation"`. I'll fix it after your review

aristov:
> To fix it we can add 'aria-hidden' attribute for group-title element.

Ok, use `aria-hidden`

aristov:
> 2nd item of #1508 description is not satisfied.

Don't forget this. Try to use `aria-activedescendant` and `aria-label` attributes to get the described behaviour.

common.blocks/menu-item/menu-item.bemhtml
```diff
@@ -2,7 +2,20 @@ block('menu-item')(
...
-    attrs()({ role : 'menuitem' }),
+    attrs()(function(){
+        var mods = this.mods,
+            menuMode = this._menuMods && this._menuMods.mode,
+            role = menuMode?
+                        (menuMode === 'check'? 'menuitemcheckbox' : 'menuitemradio') :
+                        'menuitem',
+            attrs = {
+                role : role,
+                'aria-disabled' : mods.disabled,
```
aristov, line 20: Omit `aria-disabled="false"`

common.blocks/menu-item/menu-item.bh.js
```diff
@@ -9,6 +12,10 @@ module.exports = function(bh) {
 
         ctx
             .js({ val : json.val })
-            .attr('role', 'menuitem');
+            .attrs({
+                role : role,
+                'aria-disabled' : ctx.mod('disabled'),
```
aristov, line 19: Omit `aria-disabled="false"`

common.blocks/menu-item/menu-item.deps.js
```diff
@@ -5,4 +5,10 @@
 {
     tech : 'spec.js',
     mustDeps : { tech : 'bemhtml', block : 'menu-item' }
+},
+{
+    tech : 'tmpl-spec.js',
+    mustDeps : [
+        { tech : 'bemhtml', block : 'link' }
```
aristov, line 12: What's wrong with dependencies here?

common.blocks/menu/menu.bemhtml
```diff
@@ -38,13 +38,23 @@ block('menu')(
...
         });
     }),
     attrs()(function() {
-        var attrs = { role : 'menu' };
-        this.mods.disabled || (attrs.tabindex = 0);
+        var ctx = this.ctx,
+            mods = this.mods,
+            attrs = {
+                role : 'menu',
+                'aria-disabled' : mods.disabled
```
aristov: Omit `aria-disabled="false"`. It occurs on explicit `_disabled_false` definition.

common.blocks/menu/menu.bemhtml
```diff
@@ -38,13 +38,23 @@ block('menu')(
...
+            };
+
+        mods.disabled?
+            attrs['aria-disabled'] = true :
```
aristov, line 50: Do you set `aria-disabled` attribute twice? It seems that something wrong with logic here.

common.blocks/menu/menu.bemhtml
```diff
@@ -38,13 +38,23 @@ block('menu')(
...
+            attrs.tabindex = ctx.tabIndex || 0;
```
aristov: Is `tabIndex` field to be a part of the public API now?

common.blocks/menu/menu.bh.js
```diff
@@ -1,18 +1,26 @@
...
-        };
+        var mods = ctx.mods(),
+            menuMods = {
+                theme : mods.theme,
+                disabled : mods.disabled,
+                mode : mods.mode
+            },
+            attrs = {
+                role : 'menu',
+                'aria-disabled' : mods.disabled
```
aristov: Omit `false`

common.blocks/menu/menu.bh.js
```diff
@@ -1,18 +1,26 @@
...
         ctx
             .js(true)
             .tParam('menuMods', menuMods)
             .mix({ elem : 'control' });
 
-        var attrs = { role : 'menu' };
-        ctx.mod('disabled') || (attrs.tabindex = 0);
+        mods.disabled?
+            attrs['aria-disabled'] = true :
+            attrs.tabindex = json.tabIndex || 0;
```
aristov: `tabIndex` field

common.blocks/menu/menu.bh.js
```diff
@@ -1,18 +1,26 @@
...
 
         ctx
             .js(true)
             .tParam('menuMods', menuMods)
             .mix({ elem : 'control' });
 
-        var attrs = { role : 'menu' };
-        ctx.mod('disabled') || (attrs.tabindex = 0);
+        mods.disabled?
+            attrs['aria-disabled'] = true :
```
aristov: `aria-disabled` twice.

common.blocks/menu/menu.js
```diff
@@ -37,6 +37,19 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
             this.getItems().forEach(function(menuItem){
                 menuItem.setMod(modName, modVal);
             });
+            this.domElem.attr('aria-disabled', !!modVal);
```
aristov: No `false` value here.

common.blocks/menu/menu.spec.js
```diff
@@ -90,6 +90,16 @@ describe('menu', function() {
...
         });
+
+        it('should set correct "aria-disabled" attributes on disable', function() {
+            menu
+                .setMod('disabled')
+                .domElem.attr('aria-disabled').should.be.equal('true');
+
+            menu
+                .delMod('disabled')
+                .domElem.attr('aria-disabled').should.be.equal('false');
```
aristov: `undefined`

common.blocks/menu/menu.js
```diff
@@ -135,11 +148,19 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
      */
     _onItemHover : function(item) {
         if(item.hasMod('hovered')) {
+            var id = this._activedescendant;
+
             this._hoveredItem && this._hoveredItem.delMod('hovered');
+            item.domElem.attr('id', id);
```
aristov: I'm not sure if we use `id` properly here.
cc @narqo , @dfilatov , @veged

common.blocks/menu/menu.js
```diff
@@ -37,6 +37,19 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
             this.getItems().forEach(function(menuItem){
                 menuItem.setMod(modName, modVal);
             });
+            this.domElem.attr('aria-disabled', !!modVal);
+        },
+
+        'focused' : {
+            'true' : function() {
+                this.__base.apply(this, arguments);
+                this._activedescendant || (this._activedescendant = identify());
```
narqo: 1) It looks like the code relayed on `_activedescendant` won't work properly if `menu` is initialised with `focused` modifier set.
2) As there can't be several elements on the page with `focused` modifier (we can't have several focused buttons simultaneously), it looks redundant to call `identify()` for every menu on the page. Can't we have it as a local constant? E.g. 

``` js
var ACTIVE_DESCENDAT_ID = identify()
// ···
onSetMod : {
  focused : {
    true : function() {
      this.domElem.attr('aria-activedescendant', this._activeDescendant = ACTIVE_DESCENDAT_ID);
    },
    '' : function() {
      this.domElem.removeAttr('aria-activedescendant');
      this._activeDescendant = null;
    }
```

common.blocks/menu/menu.js
```diff
@@ -135,11 +148,19 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
...
+            item.domElem.attr('id', id);
+            this.domElem.attr('aria-activedescendant', id);
             this._scrollToItem(this._hoveredItem = item);
         } else if(this._hoveredItem === item) {
+            item.domElem.removeAttr('id');
+            this.domElem.removeAttr('aria-activedescendant');
             this._hoveredItem = null;
         }
+
+        this.emit('item-hover');
```
narqo: I couldn't find the place, where you use this event.

common.blocks/menu-item/menu-item.bemhtml
sipayRT, line 20: But if block will be without `disabled` mod `'aria-disabled'` attribute will not be added. So, this attribute will be added just in case when `disabled` mod will be `true`

common.blocks/menu-item/menu-item.deps.js
sipayRT, line 12: without these deps templates for block `link` will not be applied in tmpl-specs.

common.blocks/menu-item/menu-item.bemhtml
sipayRT, line 20: If you talking about `_disabled_false` case — we don't need to support it. See [documentation](https://en.bem.info/libs/bem-components/v2.2.1/desktop/menu-item/#disabled) for details

common.blocks/menu/menu.bemhtml
sipayRT: We don't support such value - see [documentation](https://en.bem.info/libs/bem-components/v2.2.1/desktop/menu/#disabled)

common.blocks/menu/menu.bemhtml
sipayRT, line 50: yes, my fault :(

common.blocks/menu/menu.bemhtml
sipayRT: no, but we can add it to documentation. Many our blocks has this API

common.blocks/menu/menu.js
sipayRT: tnx, your variant much better

common.blocks/menu/menu.js
sipayRT: @aristov what you mean?

common.blocks/menu/menu.js
sipayRT: I added this event to use it in `select`

common.blocks/menu-item/menu-item.bh.js
sipayRT, line 19: see my answer above

sipayRT:
> Don't forget this. Try to use aria-activedescendant and aria-label attributes to get the described behaviour.

@aristov but when menu is focused no one menu-item has `hovered` state and we don't know which item should be said by JAWS. We can set `hovered` modifier for first item on menu focus, but I think it's not good idea.

aristov:
@sipayRT no `_hovered`. We should use the first `menu-item_checked` for that purpose.

sipayRT:
@aristov set id to first menu item on menu hover. it looks better now

sipayRT:
@aristov this branch was rebased and updated. Need your review

aristov:
JAWS doesn't speak `content` of the first `menu-item` of `menu-focused` (2nd item of #1508). Use `aria-label` to fix it.

sipayRT:
@aristov I just checked it one more time - on TAB event (when menu is focused) JAWS says "menu, item one not checked" and then "to move ... bla bla bla". Maybe you should clean your caches

sipayRT:
Remove `aria-checked` attribute if `menu` hasn't any `mode` modifier

sipayRT:
Clear `uniqId` on blur

sipayRT:
Try to generate id for each menu-item

sipayRT:
@aristov check now, plz

common.blocks/menu/__group/menu__group.bemhtml
```diff
@@ -6,7 +6,7 @@ block('menu').elem('group')(
         }),
         content()(function() {
             return [
-                { elem : 'group-title', content : this.ctx.title },
+                { elem : 'group-title', attrs : { 'aria-hidden' : true }, content : this.ctx.title },
```
aristov: Let's wrap this long line.

common.blocks/menu/__group/menu__group.bh.js
```diff
@@ -8,7 +8,7 @@ module.exports = function(bh) {
             ctx
                 .attr('aria-label', title, true)
                 .content([
-                    { elem : 'group-title', content : title },
+                    { elem : 'group-title', attrs : { 'aria-hidden' : true }, content : title },
```
aristov: Let's wrap this long line.

common.blocks/menu/menu.bemhtml
```diff
@@ -38,13 +38,19 @@ block('menu')(
...
-                disabled : mods.disabled
+                disabled : mods.disabled,
+                mode : mods.mode
             }
         });
     }),
     attrs()(function() {
-        var attrs = { role : 'menu' };
-        this.mods.disabled || (attrs.tabindex = 0);
+        var mods = this.mods,
```
aristov: Redundant `mods` variable.

common.blocks/menu/menu.js
aristov: Let's make this change in `select`'s PR. So we will see, how it'll be used and try to find the way to avoid it.

common.blocks/menu/menu.js
```diff
@@ -217,7 +239,22 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
...
+    _activedescendantControl : function(isFocused) {
+        var firstItemId = this.getItems()[0].domElem.attr('id');
+
+        if(isFocused) {
+            this.domElem.attr('aria-activedescendant', firstItemId);
+        } else {
+            this.domElem.removeAttr('aria-activedescendant');
+        }
     }
+
```
aristov: Redundant line break.

aristov:
I've checked the second point of #1508:
- on `menu` without `_mode` it works properly
- on `menu_mode_*` JAWS reads the first `menu-item` wrongly

I have no idea why it works so and how to fix it. Besides, @sipayRT can not reproduce this. We assume that this is a JAWS internal bug.

I think we should leave the solution, based on the permanent `id`s on `menu-item`s. It looks more proper, than move `id` from item to another.

Now all points of issue are satisfied, but there are some notes in the code changes. Fix it and ping me again, please.

sipayRT:
@aristov All your notes were fixed.

PR #1588, opened 2015-07-17
Adopt to BH 4.x

### LINKED ISSUE #1587: Support BH 4.0
Should replace `blockMods` with `mods` in
common.blocks/dropdown/_switcher/dropdown_switcher_button.bh.js
common.blocks/dropdown/_switcher/dropdown_switcher_link.bh.js
common.blocks/select/__button/select__button.bh.js
common.blocks/select/__control/select__control.bh.js

qfox:
Fixes #1587

### DIFF
--- a/common.blocks/button/_type/button_type_link.bh.js
+++ b/common.blocks/button/_type/button_type_link.bh.js
@@ -6,7 +6,7 @@ module.exports = function(bh) {
         json.target && ctx.attr('target', json.target);
         ctx.mod('disabled')?
             ctx
-                .attr('aria-disabled', true)
+                .attr('aria-disabled', 'true')
                 .js({ url : json.url }) :
             ctx.attr('href', json.url);
     });
--- a/common.blocks/button/button.bh.js
+++ b/common.blocks/button/button.bh.js
@@ -21,7 +21,7 @@ module.exports = function(bh) {
             .mix({ elem : 'control' }); // NOTE: satisfy interface of `control`
 
         if(ctx.mod('disabled')) {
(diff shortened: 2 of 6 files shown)


qfox:
Test with bh 4.0:
https://github.com/bem/bem-components/compare/v2...issues/1587%40v2-bh4
https://travis-ci.org/bem/bem-components/builds/71415623

common.blocks/button/_type/button_type_link.bh.js
```diff
@@ -6,7 +6,7 @@ module.exports = function(bh) {
         json.target && ctx.attr('target', json.target);
         ctx.mod('disabled')?
             ctx
-                .attr('aria-disabled', true)
+                .attr('aria-disabled', 'true')
```
aristov, line 9: Is it changed in BH 4.x?

qfox, line 9: In 4.1, https://github.com/bem/bh/compare/v4.0.0...v4.1.0

Boolean attributes support.

tadatuta, line 9: In readme there's still just boolean `true` not `'true'` string

qfox, line 9: https://github.com/bem/bh#ctxattrkey-value-force — I see just one example here. But behaviour have been changed. https://github.com/bem/bh/blob/master/test/test.attr.js#L29-L32 — since 4.1 it renders `hidden` instead of `hidden="true"`. And to render with `="true"` we should pass there a string.

Filed an issue: https://github.com/bem/bh/issues/158

tadatuta:
@narqo @veged @mishanga  let's merge?

veged:
:+1:

PR #1593, opened 2015-07-21
menu: remove unused deps, fix deps for tmpl-specs

### LINKED ISSUE #1562: Check if menu__group_has-title dependency is really useful
Now in `common.blocks/menu/menu.deps.js` there's `{ elems : { elem : 'group', mods : { 'has-title' : true } } }` dependency which is only referenced in two files in `bem-components`:
1. common.blocks/select/__menu/select__menu.bemhtml
2. common.blocks/select/__menu/select__menu.bh.js
and as so looks quite strange at least in `meny.deps.js` file.

sipayRT:
Closes #1562 

// @tadatuta

### DIFF
--- a/common.blocks/menu/menu.deps.js
+++ b/common.blocks/menu/menu.deps.js
@@ -5,7 +5,7 @@
         'menu-item',
         { block : 'keyboard', elem : 'codes' },
         { mods : { focused : true } },
-        { elems : { elem : 'group', mods : { 'has-title' : true } } }
+        { elem : 'group' }
     ]
 },
 {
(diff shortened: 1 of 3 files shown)


sipayRT:
@dfilatov you added modifier `'has-title'` in [select__menu.bemhtml](https://github.com/bem/bem-components/blob/v2/common.blocks/select/__menu/select__menu.bemhtml#L39)(see blame of this file). Can you explain — for what? Can we remove it?

dfilatov:
I think it used to use for css but now can be removed.

sipayRT:
tnx, will be removed

sipayRT:
@tadatuta ready

PR #1595, opened 2015-07-22
select: Don't touch dimensions of DOM elements if not necesary

narqo:
Both access to CSS `min-width` and DOM `offsetWidth` trigger Layout and Paint events in browsers. This may lead to unexpected delays during selects' initialisation.

With this changes dimensions are touched only before `menu` block is going to become visible or if `button`'s width was potentially changed.

/cc @dfilatov

### DIFF
--- a/common.blocks/select/select.js
+++ b/common.blocks/select/select.js
@@ -46,10 +46,9 @@ provide(BEMDOM.decl(this.name, /** @lends select.prototype */{
                     }, this);
 
                 this._isPointerPressInProgress = false;
+                this._buttonWidth = null;
 
                 this.hasMod('focused') && this._focus();
-
-                this._updateMenuWidth();
             }
         },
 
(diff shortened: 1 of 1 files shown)


dfilatov:
Could you add some numbers showing something?

narqo:
> Could you add some numbers showing something

Sure 600 and 1200. Just kidding :))

Here are two pages with 30 `select`s:
- http://narqo.github.io/bem-components-bench/benchs/v2/select.html (current bem-components#v2)
- http://narqo.github.io/bem-components-bench/benchs/next/select.html (version with this fixes)

Click on the "Очистить" button leads to synchronous initialisation of all `select`s via `this.findBlocksInside('select')`. See [the code](https://github.com/narqo/bem-components-bench/blob/master/blocks/form/form.js#L8-L14)

As we have a lot of "Forced  synchronous layout" calls during `onSetMod → js → inited` processing in `select`s in v2 version now, the freeze during initialisation can reach 1200ms on the mobile devices (checked with iPhone 4s).
screen shot 2015-07-23 at 13 51 11-1

dfilatov:
:+1:

Issue #1568, opened 2015-07-01
dist: Freeze dev.css bundles


tadatuta:
Images in dev version of dist bundles are not freezed now: https://github.com/bem/bem-components-dist/blob/v2/desktop/bem-components.dev.css#L202

narqo:
Don't we want to use more common naming for files, e.g. `library.js` for unminified version and `library.min.js` for minified?

tadatuta:
I think in most cases minified versions will be used anyway so it's convenient to have it as short as possible. And for debug developers just need to add `dev` suffix.

narqo:
The problem is that it's not obvious where this `.dev` suffix should be known from if user needs it. 

As we usually use `_name.js` notation for minified version and there is no `min.js` file I was sure that we just don't have dev version on CDN.

Just look at the list here https://tech.yandex.ru/jslibs/

Anyway, I don't insist.

tadatuta:
That's quite common approach but still there's no consistency there:
- backbone-min.js (minus instead of dot)
- dojo.js and dojo.js.uncompressed.js

We also checked some external libraries and found out that actually everybody choose different filenames. One company even chosen leading underscore for compressed version ;)

narqo:
Ok, now as now how to get dev version I'm fine with current approach. It was just not obvious at first sight.

aristov:
@tadatuta please specify more details in issue description.

tadatuta:
@aristov done

PR #1615, opened 2015-08-04
dist: Freeze dev.css bundles

tadatuta:
#1568


tadatuta:
gemini tests failed but the diff is not related to PR (and actually seems floating, I asked @SevInf about it)

PR #1641, opened 2015-08-18
link: update islands design

sipayRT:


tadatuta:
hovered `_view_ghost` should be black

tadatuta:
Close in favor of https://github.com/bem/bem-components/pull/1771

PR #1649, opened 2015-08-24
input: extend attributes support

alexbaumgertner:
Add `require` and `type` block's fields.
http://www.w3.org/TR/html/forms.html#the-input-element


sipayRT:
@alexbaumgertner why don't you add it on your project level? I think, these attributes not so useful, so there is no sense to add them to API

alexbaumgertner:
@sipayRT I already added.
I belive, its common case. By current `input` block you can **not** add **any** attribute to `<input>` DOM node.
I can keep only extend attributes, if you insist.
And I still think that `required`, is at least required. :)

aristov:
@alexbaumgertner these HTML-attributes should be represented as modifiers of the block.

alexbaumgertner:
@aristov I agree. I'll rewrite code :)

aristov:
@alexbaumgertner don't forget about [declared browsers support](https://github.com/bem/bem-components#supported-browsers). When you make these modifiers, you create a public API. It must work in all supported browsers.

PR #1656, opened 2015-08-26
Fix gemini test for dist

sipayRT:


sipayRT:
@blond review plz. You can try to commit something to this branch

blond:
LGTM

PR #1667, opened 2015-09-10
select should have deps icon

voischev:
Block `icon` in template inside https://github.com/bem/bem-components/blob/v2/common.blocks/select/__button/select__button.bemhtml#L19

### DIFF
--- a/common.blocks/select/__button/select__button.deps.js
+++ b/common.blocks/select/__button/select__button.deps.js
@@ -1,3 +1,3 @@
 ({
-    shouldDeps : 'button'
+    shouldDeps : ['button', 'icon']
 })


tadatuta:
@voischev 
Dependency is declared on design level: https://github.com/bem/bem-components/blob/v2/design/common.blocks/select/_theme/select_theme_islands.deps.js#L8

Are you sure it's really needed without design anyways?

voischev:
What about the right tag... & icon template in common.blocks?

voischev:
Why is there a deps on `button`, there is no deps on `icon`. Although in the template, it is also present too.

narqo:
Ok, looks fine. Thanks.

PR #1694, opened 2015-10-20
Fix select_mode_check and checkedText when single option selected

doochik:
Select must use `checkedText` option always.

Without this fix `select_mode_check` uses `text` option instead of `checkedText` when single option selected.

### DIFF
--- a/common.blocks/select/_mode/select_mode_check.js
+++ b/common.blocks/select/_mode/select_mode_check.js
@@ -32,9 +32,9 @@ provide(Select.decl({ modName : 'mode', modVal : 'check' }, /** @lends select.pr
             .toggleMod('checked', true, '', !!checkedItems.length)
             .setText(
                 checkedItems.length === 1?
-                    checkedItems[0].getText() : // one checked
+                    checkedItems[0].getCheckedText() : // one checked
                     checkedItems.reduce(function(res, item) { // many checked
-                        return res + (res? ', ' : '') + (item.params.checkedText || item.getText());
+                        return res + (res? ', ' : '') + item.getCheckedText();
                     }, '') ||
                         this.params.text); // none checked
     },
--- a/common.blocks/menu-item/menu-item.js
+++ b/common.blocks/menu-item/menu-item.js
@@ -71,6 +71,14 @@ provide(BEMDOM.decl(this.name, /** @lends menu-item.prototype */{
         return this.params.text || this.domElem.text();
     },
 
+    /**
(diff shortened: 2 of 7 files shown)


doochik:
hmmm, that's not my fault

```
> bem-components@2.3.0 test-specs /home/travis/build/bem/bem-components

> magic run specs

Error: Cannot find module 'lodash'

    at Function.Module._resolveFilename (module.js:338:15)

    at Function.Module._load (module.js:280:25)

    at Module.require (module.js:364:17)

    at require (module.js:380:17)

    at Object.<anonymous> (/home/travis/build/bem/bem-components/node_modules/enb-bem-tmpl-specs/lib/plugin.js:5:9)

    at Module._compile (module.js:456:26)

    at Object.Module._extensions..js (module.js:474:10)

    at Module.load (module.js:356:32)

    at Function.Module._load (module.js:312:12)

    at Module.require (module.js:364:17)
```

common.blocks/select/_mode/select_mode_check.js
```diff
@@ -32,9 +32,9 @@ provide(Select.decl({ modName : 'mode', modVal : 'check' }, /** @lends select.pr
             .toggleMod('checked', true, '', !!checkedItems.length)
             .setText(
                 checkedItems.length === 1?
-                    checkedItems[0].getText() : // one checked
+                    checkedItems[0].getCheckedText() : // one checked
                     checkedItems.reduce(function(res, item) { // many checked
-                        return res + (res? ', ' : '') + (item.params.checkedText || item.getText());
+                        return res + (res? ', ' : '') + item.getCheckedText();
```
just-boris, line 37: Why here `reduce` is used? Since you are using same logic for single checked item and multiple, you can simplify this code using `.map().join()`:

``` js
.setText(
  checkedItems.length === 0 ?
    this.params.text : // none checked
    checkedItems.map(function(item) {
       return item.getCheckedText();
    }).join(', ');
);
```

doochik, line 37: I don't know. It was here )
Yep this place can be simplified and it's subject for another PR

narqo:
Actually this particular behaviour is described in our designer's guide. But as it breaks some usecases, we have to wait for reply from our designers, about what should be done here.

narqo:
After person discussion it turned out, that current behaviour is described in our design guide, so "won't fix", sorry. You could implement the behaviour you need by yourself.

PR #1699, opened 2015-10-28
emit events on modal open and hide

just-boris:
It is nice to have an ability to do some actions after modal close

### DIFF
--- a/common.blocks/modal/modal.js
+++ b/common.blocks/modal/modal.js
@@ -28,6 +28,11 @@ provide(BEMDOM.decl(this.name, /** @lends modal.prototype */{
 
         'visible' : function(modName, modVal) {
             this._popup.setMod(modName, modVal);
+            if(modVal) {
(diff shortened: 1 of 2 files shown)


dfilatov:
It's absolutely redundant to trigger events which duplicate embedded events on modifier value changing, see [События при изменении модификаторов](https://ru.bem.info/technology/i-bem/v2/i-bem-js-events/#%D0%A1%D0%BE%D0%B1%D1%8B%D1%82%D0%B8%D1%8F-%D0%BF%D1%80%D0%B8-%D0%B8%D0%B7%D0%BC%D0%B5%D0%BD%D0%B5%D0%BD%D0%B8%D0%B8-%D0%BC%D0%BE%D0%B4%D0%B8%D1%84%D0%B8%D0%BA%D0%B0%D1%82%D0%BE%D1%80%D0%BE%D0%B2).

PR #1718, opened 2015-11-15
travis: Run tests on node 5

tadatuta:


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-  - "0.12"
+  - "5.0"
```
qfox, line 4: Vote for both: 4.x and 5.x, or just 4.x (not 5).

p.s. Don't use `.0`.

narqo:
@tadatuta bem-components doesn't provide anything specific to any nodejs version so why do we need this PR now?

.travis.yml
narqo, line 4: > p.s. Don't use .0.

It's definitely out of the topic, but 0.x would be widely used in [enterprise company title]'s production until Q1-Q2/2016.

tadatuta:
@narqo my first intention was to speed up building process. but now I found out at least three issues with build on `5.0`, two were fixed and tests still fall so now we're aware of these problems ;)

.travis.yml
qfox, line 4: @narqo Ehm. Not only [enterprise company title], a lot of companies, actually, will use `0.x`. But I'm not against `0.x`, I'm against sticking to `4.0`/`5.0` instead of `4.x`/`5.x`.

PR #1735, opened 2015-12-04
Fix strange deps of dropdown

### LINKED ISSUE #1731: Fix strange deps of dropdown
https://github.com/bem/bem-components/blob/v2/design/common.blocks/dropdown/_theme/dropdown_theme_simple.deps.js

Seems it should be like this:

``` js
[
    {
        shouldDeps : [
            { block : 'popup', mods : { theme : 'simple' } },
        ]
    },
    {
        include : false,
        mods : { switcher : 'link' },
        shouldDeps : { block : 'link', mod : 'theme', val : 'simple' }
    },
    {
        include : false,
        mods : { switcher : 'button' },
        shouldDeps : { block : 'button', mod : 'theme', val : 'simple' }
    }
]
```

tadatuta:
Closes #1731 

I'm not sure if it's a proper fix (or even if a bug really exists) as there's no support for `include: false` in ENB anyways.

cc @mishanga @veged

### DIFF
--- a/design/common.blocks/dropdown/_theme/dropdown_theme_simple.deps.js
+++ b/design/common.blocks/dropdown/_theme/dropdown_theme_simple.deps.js
@@ -1,15 +1,17 @@
-({
-    shouldDeps : [
-        { block : 'popup', mods : { theme : 'simple' } },
-        {
-            include : false,
-            mods : { switcher : 'link' },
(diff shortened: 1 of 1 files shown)


mishanga:
13836689328754_yapfiles_ru

Я не знаю, как правильно, тебе виднее.

qfox:
Первое явно кривое. :-)

PR #1740, opened 2016-01-12
Fixed left border of button in control group in old IE.

GreLI:
Same problem as in #1723 but in older IE.

Before:
leftborderie8-before

After:
leftborderie8-after


tadatuta:
:+1:

Issue #1734, opened 2015-12-04
select: Fix a11y markup


tadatuta:
Current `select` HTML generated with `bem-xjst` looks as follows (note `uniq*` occurrences):

``` html
<div class="select select_mode_radio-check select_theme_islands select_size_m i-bem" data-bem='{"select":{"name":"select3","text":"—"}}'>
    <input class="select__control" type="hidden" name="select3" value="2" />
    <button class="button button_size_m button_theme_islands button_checked button__control select__button i-bem" data-bem='{"button":{}}' role="listbox" type="button" aria-owns="uniq14492417126291 uniq14492417126291 uniq14492417126291" aria-labelledby="uniq14492417126292">
        <span class="button__text" id="uniq14492417126292">Мастер-класс</span>
        <span class="icon select__tick"></span>
    </button>
    <div class="popup popup_target_anchor popup_theme_islands popup_autoclosable i-bem" data-bem='{"popup":{"directions":["bottom-left","bottom-right","top-left","top-right"]}}' aria-hidden="true">
        <div class="menu menu_size_m menu_theme_islands menu_mode_radio-check menu__control select__menu i-bem" data-bem='{"menu":{}}'>
            <div class="menu-item menu-item_theme_islands i-bem" data-bem='{"menu-item":{"val":1}}' role="option" id="uniq14492417126293" aria-checked="false">Доклад</div>
            <div class="menu-item menu-item_checked menu-item_theme_islands i-bem" data-bem='{"menu-item":{"val":2}}' role="option" id="uniq14492417126294" aria-checked="true">Мастер-класс</div>
            <div class="menu-item menu-item_theme_islands i-bem" data-bem='{"menu-item":{"val":3}}' role="option" id="uniq14492417126295" aria-checked="false">Круглый стол</div>
        </div>
    </div>
</div>
```

`BH` results with different ids in `aria-owns` attribute but still unique across markup which seems to be strange.

tadatuta:
cc @sipayRT @aristov

veged:
@tadatuta could you describe what exactly "strange" is? we looked at https://github.com/bem/bem-components/pull/1743/files and could't catch the point

tadatuta:
As it goes in description, there's a bug in uniq id generation so the same id is generated for `aria-owns` (`"uniq14492417126291 uniq14492417126291 uniq14492417126291"` in example above)

veged:
ok, I got it and just add some comment to PR

aristov:
Fixed

PR #1743, opened 2016-01-25
select: Fix a11y markup

aristov:


aristov:
#1734

tadatuta:
There's codestyle error in PR

aristov:
> There's codestyle error in PR

Where?

tadatuta:
https://travis-ci.org/bem/bem-components/jobs/104734867#L592

common.blocks/select/select.bemhtml
```diff
@@ -21,7 +21,7 @@ block('select')(
                         iterateOptions(option.group);
                     } else {
                         firstOption || (firstOption = option);
-                        optionIds.push(option.id = _this.generateId());
+                        optionIds.push(option.id = 'option' + random++);
```
veged: instead of `'option' + random++` (with `random` variable) we can just use `_this.generateId() + 'option' + i`

aristov: If we have `group`, the `i` iterator will starts from zero. `id` attribute will not be unique in this case.

veged: yeah... so even better, you can use `option.id = _this.identify(option)`!

aristov: What about `bh`? There is no `identify` method.

veged: as I can see in [code](https://github.com/bem/bh/blob/dff393fa5c6c7f1d0a327a1c2f43895bcd23a0c7/lib/bh.js#L256) and [specs](https://github.com/bem/bh/blob/dff393fa5c6c7f1d0a327a1c2f43895bcd23a0c7/test/test.generateId.js#L10) `generateId` in BH works different (also incorrect) and generate different ids every time — so we can just do `option.id = ctx.generateId()`

aristov: Done.

veged:
now changes looks good for me —@aristov, can you rebase and merge?

narqo:
@aristov for future works, please, squash your commits, before merging, so PR'd be more solid.

PR #1751, opened 2016-02-16
Update npm deps

tadatuta:
---

Showcase is [available](http://bem.github.io/reports/5956/showcase/showcase.html)


tadatuta:
@birhoff В этом PR упали gemini-тесты, но репорт не был закоммичен. Посмотри, пожалуйста.

package.json
```diff
@@ -15,41 +15,41 @@
     "node": ">=0.10"
   },
   "dependencies": {
-    "bem-environ": "~1.4.0"
+    "bem-environ": "^1.4.0"
```
narqo: Do we really need this crap?

tadatuta: we may drop quite a few deps if we get rid of bem-tools in a lib but as far as I know some people use it for `bem create` command

PR #1760, opened 2016-03-01
Fixed links in showcase

GreLI:
In [showcase](http://bem.github.io/reports/5748/showcase/showcase.html) links are going to

> http://bem.github.io/reports/common.blocks/button/button.ru.md#Модификатор-view

which is 404. There is no `common.blocks` in [bem/reports](https://github.com/bem/reports). I believe it was intended to link to urls like

> https://github.com/bem/bem-components/blob/v2/common.blocks/button/button.ru.md


tadatuta:
@GreLI thank you!

PR #1761, opened 2016-03-01
Fix gemini reports

birhoff:
---

Showcase is [available](http://bem.github.io/reports/5757/showcase/showcase.html)


tadatuta:
Fixed in https://github.com/bem/bem-components/commit/7674f59aafd16fb13ee2c1cb4c182334f35f34de

Issue #1768, opened 2016-03-21
checkbox, radio: Sync prop to support jQuery form serialization


tadatuta:
Now just first time checkbox/radio is check it works properly. Since the second one form serialization will fail.

PR #1769, opened 2016-03-21
checkbox, radio: Sync prop to support jQuery form serialization #1768

tadatuta:
---

Showcase is [available](http://bem.github.io/reports/5807/showcase/showcase.html)

### DIFF
--- a/common.blocks/checkbox/_type/checkbox_type_button.spec.js
+++ b/common.blocks/checkbox/_type/checkbox_type_button.spec.js
@@ -37,6 +37,19 @@ describe('checkbox_type_button', function() {
             button.hasMod('checked').should.be.false;
         });
 
+        it('should set/unset property properly', function() {
+            expect(checkbox.elem('control')[0].checked).to.be.false;
+
+            checkbox.setMod('checked');
+            expect(checkbox.elem('control')[0].checked).to.be.true;
+
+            checkbox.delMod('checked');
+            expect(checkbox.elem('control')[0].checked).to.be.false;
+
+            checkbox.setMod('checked');
+            expect(checkbox.elem('control')[0].checked).to.be.true;
+        });
+
         it('should set/unset aria-attributes properly', function() {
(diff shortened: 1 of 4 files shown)


tadatuta:
cc @dfilatov

dfilatov:
Please, add more information. I don't understand what you mean by "subsequent checks".

tadatuta:
@dfilatov  Way to reproduce on current `v2 HEAD`:
1. Check `checkbox_type_button` // everything is ok
2. Uncheck // still ok
2. Check again // since this moment form won't serialize this checkbox anymore

narqo:
Could the issue be reproduced using API calls? If so, tests are strongly advisable as this looks like another degradation connected with a11y refactoring.

tadatuta:
@narqo I added test, it fails with old code and passes with my changes, but I'm not sure about implementation, though. Please take a look.

narqo:
As I can see from the [implementation of `$.serializeArray()`](https://github.com/jquery/jquery/blob/70526981916945dc4093e116a3de61b1777d4718/src/serialize.js#L108), we should have the same "issue" in `radio`, as it uses `attr/removeAttr` as well (one could check some other blocks, I don't remember all the details of our a11y refactoring).

Just to be noted: I believe the title of issue/pr should be change because it's about jQuery's implementation of it's serialiser or even about how jQuery deal with internal DOM properties.

It was obvious that `attr('checked')` and `removeAttr('checked')` don't do anything with internal DOM properties, but only deal with HTML attributes. Do we expect this DOM APIs to work? If yes, the test might be significantly simplified to something like this

``` js
checkbox.setMod('checked');
checkbox.elem('control')[0].checked.should.be.true;
```

dfilatov:
@tadatuta Could you fix the same stuff in `radio` as well and simplify specs?

tadatuta:
@dfilatov, sure, will do it today

Issue #1703, opened 2015-11-10
Focus should not enlarge components


tadatuta:
E.g. `input` in focus should be of the same height as without it. Same thing for any other block.

vkhv:
Have you meant it? 

![input](http://i.imgur.com/6wYbDge.png)

vkhv:
i can offer it - https://github.com/bem/bem-components/pull/1742

PR #1771, opened 2016-03-24
Update design

gela-d:
@tadatuta

---

Showcase is [available](http://bem.github.io/reports/6001/showcase/showcase.html)

Report is [available](http://bem.github.io/reports/6001/gemini-report)


narqo:
1) Could we add focus style for `link` as well? See #1223 for the final style.
2) `button_view_action` constantly pulsing after being clicked — not sure this behaviour is right.
3) What about this #1703 bug? Is it not valid anymore? /cc @tadatuta

tadatuta:
#1703 is still valid but was considered major and all the other changes may be minor (you'll know after @dfilatov try the branch on https://n.maps.yandex.ru/)

gela-d:
> 1) Could we add focus style for link as well? See #1223 for the final style.

@narqo Yes, I can.
There is one question about focused states.
In normal state of focus we have this screen


But if we press another link without focus, It will be focused too.
Is this correct behavior? In this state we have on link this classes `link link_theme_islands link__control i-bem islands-no-url link_js_inited link_focused link_hovered`

This is old vs new screenshots

gela-d:
> 2) button_view_action constantly pulsing after being clicked — not sure this behaviour is right.

This is focused state. After pressing this button leaves in this state, thats why it blinking.

dfilatov:
> This is focused state. After pressing this button leaves in this state, thats why it blinking.

@gela-d To prevent blinking after click, the animation should be applied for `focused-hard` state, not `focused`.

gela-d:
@narqo @dfilatov Now there is no implementation for `focused-hard` in link. Thats why I cant implement focus for link in this PR.

narqo:
> There is one question about focused states [...]

I think this is ok.

> [...] implementation for focused-hard in link.

`focused-hard` should be used to fix `button`, see the 2nd point of my comment https://github.com/bem/bem-components/pull/1771#issuecomment-200982777 above.

gela-d:
@narqo sorry, I incorrectly read your comment last time.
Update PR - https://github.com/bem/bem-components/pull/1771/commits/54ede76339add4667e5fc5fb1aa01ba7723b2970

design/common.blocks/link/_theme/link_theme_islands.styl
```diff
@@ -18,6 +18,21 @@
...
+        {
+            color: #999;
+        }
+
+        &_black
+        {
+            color: #000;
+        }
+
+        &_strong
```
dfilatov, line 37: Where do these names (`ghost`, `black`, `strong`) come from? They don't look too semantic.
Do we really need them?

tadatuta, line 37: https://nda.ya.ru/3RSDX5

narqo, line 37: But do we need them? It's trivial to implement them as well as links with other colours or font styles.

tadatuta, line 37: It's a way to keep everything consistent (see posts about different colors in header at some small search company ;))

narqo, line 37: I believe such things must be solved with proper design guides that based on strict and clear colour schemes, not by weird CSS class names, like `p_color_blue`.

Apart from discussions, current implementation doesn't allow to compose `_external` and `_strong` links, while this combination is used in the company you've been talking about ;)

gela-d:
Close via #1838

PR #1791, opened 2016-04-06
menu: After disabled state tabindex has not been set

### LINKED ISSUE #1767: menu: After disabled state tabindex has not been set
If I have menu with `disabled` mod, and then make `menu.delMod('disabled')` tabindex has not been set.
In template we have this construction https://github.com/bem/bem-components/blob/v2/common.blocks/menu/menu.bemhtml#L45-L47.

narqo:
fix #1767

/cc @dfilatov

---

Showcase is [available](http://bem.github.io/reports/5841/showcase/showcase.html)

### DIFF
--- a/common.blocks/control/control.js
+++ b/common.blocks/control/control.js
@@ -32,7 +32,10 @@ provide(BEMDOM.decl(this.name, /** @lends control.prototype */{
                     // if block already has focused mod, we need to focus control
                     this.hasMod('focused') && this._focus();
 
-                this._tabIndex = this.elem('control').attr('tabindex');
+                this._tabIndex = typeof this.params.tabIndex !== 'undefined'?
+                    this.params.tabIndex :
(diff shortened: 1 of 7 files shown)


tadatuta:
tmpl specs should be updated with this PR

narqo:
@tadatuta I'm not sure I understand, what has happened with CI here. Could you explain, what's wrong with the specs?

tadatuta:
@narqo build scrips were fixed in https://github.com/bem/bem-components/commit/a674f8475e86002b6e6e62e71d0338367a7cf02d

So I rebased the branch and added a fix for codestyle. Guess commits should be squashed before merge.

narqo:
Amazing!

tadatuta:
@narqo any reason not to merge?

narqo:
It seems that tests for dist are not stable. But after I've restart the build everything looks fine.

Issue #1741, opened 2016-01-16
control-group: Wrong styling for dropdown


tadatuta:
Since `dropdown` is two DOM nodes now (trigger + popup) `last-child` selector works wrong:
http://i.imgur.com/DAMqGDb.png

narqo:
Looks like this is a degradation after #1392.

veged:
@tadatuta can it be fixed in CSS or it's better to rollback #1392

tadatuta:
it can with several selectors with `:last-of-type`

veged:
ok, so keep it _[ready]_

narqo:
@tadatuta @gela-d any estimates here?

PR #1835, opened 2016-04-29
control-group: Wrong styling for dropdown (https://github.com/bem/bem-components/issues/1741)

vithar:


design/common.blocks/button/_theme/button_theme_islands.styl
```diff
@@ -344,8 +344,8 @@ ctx = '.button_theme_islands'
         }
     }
 
-    > {ctx}:first-child,
-    > :first-child {ctx}
+    > {ctx}:first-of-type,
```
tadatuta: looks like there's no need for this change

PR #1836, opened 2016-04-30
control-group: Wrong styling for dropdown #1741

vithar:


vithar:
https://github.com/bem/bem-components/issues/1741

PR #1838, opened 2016-05-10
Update design

gela-d:
---

Showcase is [available](http://bem.github.io/reports/6051/showcase/showcase.html)

Report is [available](http://bem.github.io/reports/6051/gemini-report)


dfilatov:
I've tried it out on nmaps. Excepting https://github.com/bem/bem-components/pull/1838/files#diff-6dce689358f9ff2e55d919fc1ad4ba16R8, everything looks good.

tadatuta:
@dfilatov is it fine for you to override it back on project level?

dfilatov:
@tadatuta I think it's a suitable solution.

dfilatov:
But don't forget about some unsemantic modifiers like "black".

gela-d:
@tadatuta This commit broke some controls. For example, 

We should look on this change carefully

dfilatov:
@gela-d good catch. I see that it has different appearance in FF and Chrome. It should be fixed.

dfilatov:
Is there any progress?

gela-d:
> Is there any progress?

Yes, we trying to fix and understand a problem with `line-heigth` form page. There is some problems with it.

gela-d:
Close via #1853

PR #1840, opened 2016-05-10
Changelog for 3.0.0

tadatuta:
---

Showcase is [available](http://bem.github.io/reports/6017/showcase/showcase.html)

Report is [available](http://bem.github.io/reports/6017/gemini-report)


tadatuta:
cc @veged @dfilatov @narqo

veged:
LGTM

PR #1854, opened 2016-05-24
packaging: Provide reasonable ignores

### LINKED ISSUE #1844: Add proper ignores for npm/bower releases

tadatuta:
As we now publish releases with precompiled stylus files to `npm` and `bower` there was request to get rid of all useless files.

Closes #1844

Based on
- https://docs.npmjs.com/files/package.json
- https://github.com/bower/spec/blob/master/json.md

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,14 +1,26 @@
 {
   "name": "bem-components",
-  "version": "3.0.0",
   "description": "BEM Components Library",
-  "bugs": "https://github.com/bem/bem-components/issues",
+  "repository": {
+    "type": "git",
+    "url": "git://github.com/bem/bem-components.git"
+  },
+  "homepage": "https://bem.info/libs/bem-components/",
   "main": "",
+  "keywords": [
+    "bem",
+    "components",
+    "controls",
+    "ui"
+  ],
   "ignore": [
-    ".bem/cache",
-    ".enb/tmp",
-    "node_modules",
-    "libs"
+    "*",
+    "!common.blocks",
+    "!desktop.blocks",
+    "!touch.blocks",
+    "!design",
+    "!*.md",
+    "!LICENSE.txt"
   ],
   "dependencies": {
     "bem-core": "3.0.1"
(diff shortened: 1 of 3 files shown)


tadatuta:
cc @veged @narqo @dima117

narqo:
If we have a non-bundled version on npm, why there is no proper dependencies in `package.json`? May be we should have at least a `peerDependency` from bem-core?

tadatuta:
@narqo sounds reasonable, will add it

bower.json
```diff
@@ -1,21 +1,32 @@
 {
   "name": "bem-components",
-  "version": "3.0.0",
```
zxqfox: Where are this disappeared?

package.json
```diff
@@ -4,17 +4,31 @@
...
   ],
   "engines": {
     "node": ">=4"
   },
+  "files": [
+    "common.blocks",
+    "desktop.blocks",
+    "touch.blocks",
+    "design"
+  ],
```
qfox, line 28: Would you mind to move tests, gemini-s, etc out of the package with `.npmignore`?

qfox:
The rest looks good to me.

package.json
```diff
@@ -4,17 +4,31 @@
...
-    "bem"
+    "bem",
+    "components",
+    "controls",
+    "ui"
   ],
   "engines": {
     "node": ">=4"
   },
+  "files": [
```
Yeti-or, line 22: are you sure that `README.ru.md` will be included by default?

qfox, line 22: Yep. https://docs.npmjs.com/files/package.json#files

Yeti-or, line 22: ~~yeah your are right just checked on npm/npm test cases~~
~~https://github.com/npm/npm/blob/master/test/tap/files-and-ignores.js#L447~~

Yeti-or, line 22: Not as easy as it could be: https://github.com/npm/npm/pull/12862

qfox, line 22: Better to add it by self, agreed.

Issue #1858, opened 2016-05-27
overused `applyNext()` calls


Guria:
In some places looks overused, because it much simpler just to pass values in ctx.
- https://github.com/bem/bem-components/blob/99ba4f447e762b76992adec927649fd74686beeb/common.blocks/input/input.bemhtml.js#L5
- https://github.com/bem/bem-components/blob/99ba4f447e762b76992adec927649fd74686beeb/common.blocks/input/__control/input__control.bemhtml.js

Is it intended or subject to fix?

miripiruni:
In first link `applyNext(…)` reasonable.
In second link there is no `applyNext()` at all.

In general `applyNext()` is OK. Why you open this task? Can we close it?

PR #1859, opened 2016-05-27
input: remove overused applyNext() call

Guria:
see #1858

### DIFF
--- a/common.blocks/input/__control/input__control.bemhtml.js
+++ b/common.blocks/input/__control/input__control.bemhtml.js
@@ -2,7 +2,7 @@ block('input').elem('control')(
     tag()('input'),
 
     attrs()(function() {
-        var input = this._input,
+        var input = this.ctx._input,
             attrs = {
                 id : input.id,
                 name : input.name,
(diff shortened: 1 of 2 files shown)


tadatuta:
@miripiruni @veged, what do you think about some formal guide about tunneling parent data to children nodes?

veged:
@tadatuta of course that's will be greater!

miripiruni:
1. There is three elements that I saw: `__box`, `__clear`, `__control`. No one uses `this._input` except `control`. Why you need tunneling all `this.ctx` in `_input` field? Moreover, it’s not good case where user create `_input` field in BEMJSON and will be very surprised at the results.
2. If the reason of refactoring is performance, why you not just simplify BEMHTML to one single template with out `applyNext()`, `def()` and any tunneling?

``` js
block('input')(
    tag()('span'),
    js()(true),
    content()(function() {
        var ctx = this.ctx,
            controlAttrs = {
                id : ctx.id,
                name : ctx.name
            };
        // …
        return {
            elem: 'box',
            content: { elem: 'control', tag: 'input', attrs: controlAttrs }
        };
    })
);
```

PR #1860, opened 2016-05-27
attach: Add cursor pointer

### LINKED ISSUE #1857: attach: cursor pointer
Cursor doesn't change to pointer when mouse over attach button, because it's covered with absolutely positioned transparent `<input type="file">`.
So the fix is just add `.attach__control { cursor: pointer }` property.

Guria:
closes #1857


tadatuta:
@Guria thank you!

PR #1881, opened 2016-07-20
Fix decl mod for progressbar

belozer:


tadatuta:
@belozyorcev thank you!

PR #1913, opened 2016-08-22
Force blur on disable to prevent "document.activeElement = button_disabled" in FF

dfilatov:
cc @narqo @tadatuta 

FF keeps disabled button as `document.activeElement` if it was focused before.

---

Showcase is [available](http://bem.github.io/reports/6241/showcase/showcase.html)

Report is [available](http://bem.github.io/reports/6241/gemini-report)

### DIFF
--- a/common.blocks/control/control.js
+++ b/common.blocks/control/control.js
@@ -100,9 +100,9 @@ provide(BEMDOM.decl(this.name, /** @lends control.prototype */{
     },
 
     _blur : function() {
-        dom.isFocusable(this.elem('control'))?
-            this.elem('control').blur() :
-            this._onBlur();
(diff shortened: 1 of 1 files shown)


dfilatov:
~~Don't merge it. WIP.~~

narqo:
Looks like a crutch™

Ok.

dfilatov:
> Looks like crutch™

I'll be glad if you provide a better alternative )

narqo:
@tadatuta could you help us with gemini and "dist" tests here? As far as I can see from the log,  tests for `spin` fail while the block isn't affected by the PR. And there is no link to tests' report in the log.

tadatuta:
@narqo trying to fix it right now, see https://github.com/bem/bem-components/pull/1914

PR #1922, opened 2016-08-29
Modal decl fix

aristov:
`decl => declMod`

---

Showcase is [available](http://bem.github.io/reports/6274/showcase/showcase.html)

Report is [available](http://bem.github.io/reports/6274/gemini-report)


aristov:
@tadatuta

PR #1953, opened 2016-10-25
button: Don't render element for text if passed `null` or `undefined`

dfilatov:
---

Report is [available](http://bem.github.io/reports/6394/gemini-report)


narqo:
@tadatuta could you check tests?

tadatuta:
merged via rebase on `v4`

PR #1963, opened 2016-11-18
modal_autoclosable: subscribe to `pointerclick` on next tick to prevent being closed immediately

dfilatov:
---

Showcase is [available](http://bem.github.io/reports/6427/showcase/showcase.html)


dfilatov:
cc @veged @narqo

veged:
👍

narqo:
OK!

PR #1969, opened 2016-11-23
menu: Correct event to live initialization

gela-d:


common.blocks/menu/menu.js
```diff
@@ -235,7 +235,7 @@ provide(BEMDOM.decl({ block : this.name, baseBlock : Control }, /** @lends menu.
 }, /** @lends menu */{
     live : function() {
         this
-            .liveInitOnBlockInsideEvent({ modName : 'hovered', modVal : '*' }, 'menu-item', function(e) {
+            .liveInitOnBlockInsideEvent({ modName : 'hovered', modVal : true }, 'menu-item', function(e) {
```
dfilatov, line 238: How will "unhover" work after this?

PR #2025, opened 2017-03-20
Drop ENB cache on libs update

tadatuta:
---

Showcase is [available](http://bem.github.io/reports/6640/showcase/showcase.html)


veged (commented):
`npm i` too?

tadatuta:
⬆️

PR #2051, opened 2017-06-20
popup_target_anchor: Change calc logic

gela-d:
In new version of jquery changes calc logic. That's why we should use clear pixels
@tadatuta


tadatuta (approved):

gela-d:
@tadatuta If all ok, cal I merge?

PR #2054, opened 2017-06-21
Update bem-core to 4.2.1

tadatuta:


tadatuta:
all the tests passed (see logs) but travis for some reason was not able to send notification

PR #2064, opened 2017-07-28
menu__item: Fix arrow/pointer highlight behavior

Yeti-or:
before:
highlight menu__item with cursor then use arrows then move the cursor again —  no highlight
menu__item_before

after:
menu__item_after


tadatuta:
@Yeti-or lgtm but there's codestyle error

Yeti-or:
yeah. And I suppose it's in dev

Yeti-or:
I mean v6

Yeti-or:
fixed id here: https://github.com/bem/bem-components/pull/2066

tadatuta:
@Yeti-or thank you!

PR #2076, opened 2018-01-26
[README.md]: Fix links

godfreyd:


tadatuta:
Specs are broken because of https://github.com/bem/bem-components/issues/2078

PR #2083, opened 2018-04-07
dropdown: Fix tmpl specs

### LINKED ISSUE #2078: Fix tmpl specs
Since https://github.com/bem/bem-xjst/releases/tag/v8.7.1 nested mixes are supported so tmpl specs for `dropdown` are broken for `bem-xjst`. And it's not enough to just update HTML as `bh` just don't have support for nested mixes.

tadatuta:
Closes #2078


tadatuta:
Since https://github.com/bem/bem-xjst/releases/tag/v8.7.1 nested mixes are supported so tmpl specs for `dropdown` are broken for `bem-xjst`.
And it's not enough to just update HTML as `bh` just doesn't have support for nested mixes.

In this PR I manually add common dropdown mixes to `dropdown__popup` (before they were added to `dropdown__switcher` only). It seems quite reasonable and I have no idea why it was not implemented this way initially.

**But I still can't decide if we can publish such update as minor**: there could be some styles for such mixes on projects, etc.

// cc @veged

PR #2092, opened 2018-04-21
Introduce .bemrc

tadatuta:


belozer:
are you planning to add to previous versions?

tadatuta:
@belozer ok, will backport it as soon as we decide for final version

tadatuta:
@zxqfox does it look good to you?

PR #2093, opened 2018-05-08
Docs: image html content example

miripiruni:


common.blocks/image/image.en.md
```diff
@@ -105,7 +105,17 @@ Specifies the image height in pixels.
 
 Type: `BEMJSON`.
 
-Specifies content of the image in SVG format.
+Specifies content of the image in SVG format. Depends on `escapeContent` option
```
tadatuta: Specifies content of the image in SVG format.

Usage depends on `escapeContent` option of `bem-xjst`.

For `escapeContent: true` (default behavior) it should be
```js
// ...
```

For `escapeContent: false` it should be
```js
// ...
```

miripiruni: @tadatuta updated.

common.blocks/image/image.ru.md
```diff
@@ -107,6 +107,6 @@
 ```js
 {
     block : 'image',
-    content : '<svg xmlns="http://www.w3.org/2000/svg" width="15" height="16"><path d="M13.5.5l-8 12L1.7 8l-1 1.6L5.6 15l9.1-13.4z"/></svg>'
+    content : { html: '<svg xmlns="http://www.w3.org/2000/svg" width="15" height="16"><path d="M13.5.5l-8 12L1.7 8l-1 1.6L5.6 15l9.1-13.4z"/></svg>' }
```
miripiruni, line 110: `content: { html: '…' }` for all cases is valid.

Issue #2094, opened 2018-06-22
move lazyInit from js to bemhtml/bh in modifers


belozer:
lazyInit in mods breake lazyInit on all blocks https://github.com/bem/bem-core/issues/1576
and not allowed after https://github.com/bem/bem-core/pull/1579/files

**upd**
or not need https://github.com/bem/bem-core/pull/1584

PR #2099, opened 2018-07-23
Update bem-core to 4.3.1

tadatuta:


belozer:
close https://github.com/bem/bem-components/issues/2094

